PR #1, opened 2019-11-21
bootstrap project

BusyJay:
- Add serveral important traits
- Fix license and README with updated name


src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+///   handle -> pause
+///     |
+///    end
+/// ```
+/// 
+/// Generally users should use the provided future thread pool or callback
+/// thread pool instead. This is only for advance customization. 
+pub trait Runner {
+    /// The task runner can handle.
+    type Task;
```
sticnarf: I once met some trouble if `Task` knows nothing about the queue (`LocalSpawn` and `RemoteSpawn` here).

For example, a callback handle should be aware of what `LocalSpawn` is, but with this trait it's hard to do that.

sticnarf (approved):
LGTM

src/runner.rs
```diff
@@ -0,0 +1,71 @@
...
+pub trait Runner {
+    /// The task runner can handle.
+    type Task;
+
+    /// The local spawn that can be accepted to spawn tasks.
+    type Spawn: LocalSpawn<Task = Self::Task>;
+
+    /// Called when the runner is started.
+    ///
+    /// It's guaranteed to be the first method to called before anything else.
```
AndreMouche: to call or to be called?

AndreMouche (commented):
Rest LGTM

src/runner.rs
BusyJay: My mistake.

AndreMouche (approved):
LGTM

PR #2, opened 2019-11-21
add callback task and runner

sticnarf:
Add callback task and its runner.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,8 +1,8 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
-/// In Yatp model, any pieces of logic aims to be executed in thread pool is
-/// called Task. There can be different definitions of Task. Some people may
-/// choose `Future` as Task, some may just want callbacks, or even Actor
+/// In the model of yatp, any piece of logic aiming to be executed in a thread
+/// pool is called Task. There can be different definitions of Task. Some people
+/// may choose `Future` as Task, some may just want callbacks, or even Actor
 /// messages. But no matter what a Task is, there should be some role know how
 /// to execute it. The role is call `Runner`.
 ///
@@ -52,14 +52,14 @@ pub trait Runner {
     fn end(&mut self, _spawn: &mut Self::Spawn) {}
 }
 
-/// Allows spawn a task to the thread pool from a different thread.
+/// Allows spawning a task to the thread pool from a different thread.
 pub trait RemoteSpawn: Sync + Send {
     type Task;
 
     fn spawn(&self, t: impl Into<Self::Task>);
 }
 
-/// Allows spawn a task inside the thread pool.
+/// Allows spawning a task inside the thread pool.
 pub trait LocalSpawn {
     type Task;
     type Remote: RemoteSpawn;
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -36,7 +36,12 @@ pub trait Runner {
     /// It's possible that a task can't be finished in a single execution, in
     /// which case feel free to spawn the task again and return false to
     /// indicate the task has not been finished yet.
-    fn handle(&mut self, spawn: &mut Self::Spawn, task: Self::Task) -> bool;
+    fn handle(
+        &mut self,
+        spawn: &mut Self::Spawn,
+        task: Self::Task,
+        ctx: &<Self::Spawn as LocalSpawn>::TaskContext,
+    ) -> bool;
 
     /// Called when the runner is put to sleep.
     fn pause(&mut self, _spawn: &mut Self::Spawn) -> bool {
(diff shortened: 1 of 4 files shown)


src/task/callback.rs
```diff
@@ -0,0 +1,237 @@
...
+pub struct Handle<'a, Spawn> {
+    spawn: &'a mut Spawn,
+    rerun: bool,
+}
+
+impl<'a, Spawn> Handle<'a, Spawn>
+where
+    Spawn: LocalSpawn<Task = Task<Spawn>>,
+{
+    /// Spawns a [`FnOnce`] to the thread pool.
```
BusyJay: Hmm, should "Fn" pronounce like "F-N", "fun" or "function"? /cc @Hoverbear @nrc @brson

src/task/callback.rs
```diff
@@ -0,0 +1,237 @@
...
+//! A [`FnOnce`] or [`FnMut`] closure.
+
+use crate::LocalSpawn;
+
+use std::marker::PhantomData;
+
+/// Tasks that is supported by [`Runner`].
+///
+/// Anything that can be converted to a [`Task`] is supported by [`Runner`].
+pub trait CallbackTask<Spawn> {
```
BusyJay: Is it still necessary given that the queue definition is hidden behind `*Spawn` trait?

src/task/callback.rs
Hoverbear: I always thought of it as "fun" 🤣I think it's as you please.

nrc: I say something like "fun" but without the "u" similar to std

src/task/callback.rs
sticnarf: The problem is that `LocalSpawn` sometimes needs extra information besides the original task.

For example, if we want to spawn a new task to the `LocalSpawn` of a multi-level queue, at least a counter recording the task's running time needs to be passed in along with the task.

sticnarf: I remove this trait and add a `TaskContext` associated type to `LocalSpawn`. I think now it's flexible enough.

Resemble to `TaskContext`, I also add `SpawnOption` to `RemoteSpawn`. Currently in my mind, it can be used to specify the priority (nice) of a task.

src/runner.rs
```diff
@@ -52,20 +57,46 @@ pub trait Runner {
     fn end(&mut self, _spawn: &mut Self::Spawn) {}
 }
 
-/// Allows spawn a task to the thread pool from a different thread.
+/// Allows spawning a task to the thread pool from a different thread.
 pub trait RemoteSpawn: Sync + Send {
     type Task;
+    type SpawnOption;
```
BusyJay, line 63: Why choose "Option" for `Remote` and "Context" for `Local`?

sticnarf: I wanted to express `Option` is something that is usually controllable by a user while `Context` is not set by the user but comes from inside the queue in most cases.

src/task/callback.rs
```diff
@@ -0,0 +1,265 @@
...
+            Task::new_once(move |_| {
+                tx.send(42).unwrap();
+            }),
+            &(),
+        );
+        assert_eq!(rx.recv().unwrap(), 42);
+    }
+
+    #[test]
+    fn test_mut_no_spin() {
```
BusyJay: I think spinning is the reason why `spawn_times` is 0.

src/task/callback.rs
```diff
@@ -0,0 +1,265 @@
...
+/// to be rerun. `max_inspace_spin` is the maximum times a task is rerun at once
+/// before being put back to the thread pool.
+pub struct Runner<Spawn> {
+    max_inplace_spin: usize,
+    _phantom: PhantomData<Spawn>,
+}
+
+impl<Spawn> Runner<Spawn> {
+    /// Creates a new runner with given `max_inplace_spin`.
+    pub fn new(max_inplace_spin: usize) -> Self {
```
BusyJay, line 103: What if `max_inplace_spin` is 0?

src/task/callback.rs
sticnarf: I should have changed the test name to `test_mut_no_respawn` and `test_mut_respawn`...

src/task/callback.rs
sticnarf: It's the **rerun** times per the doc comment. `max_insplace_spin` is 0 means never rerun inplace.
Previous implementation is incorrect. I fix it in the latest commit.

src/task/callback.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// Sets `max_inplace_spin`.
+    pub fn set_max_inplace_spin(&mut self, max_inplace_spin: usize) {
+        self.max_inplace_spin = max_inplace_spin;
+    }
+}
+
+impl<Spawn> Default for Runner<Spawn> {
+    fn default() -> Self {
+        Runner {
+            max_inplace_spin: 3,
```
BusyJay, line 119: Why use 3 instead of 4? How does it perform?

BusyJay (approved):
rest LGTM

BusyJay (commented):
rest LGTM

src/task/callback.rs
sticnarf: I just keep the behavior the same as the original code. I guess the default value is by rule of thumb and highly workload related (for callback tasks). Maybe it's not that important.

BusyJay: But it's 4 in the original code: https://github.com/BusyJay/chocolates/blob/dev/src/thread_pool/callback.rs#L89

sticnarf: 4 in that code is the same as 3 here because the value in that code is the maximum run times and here is rerun times.

BusyJay (approved):

AndreMouche (approved):
LGTM

PR #4, opened 2019-11-25
Add thread pool builder

BusyJay:
This PR adds the interfaces for thread pool.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -1,0 +1,24 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+/// A Task queue for thread pool.
+///
+/// Unlike a general MPMC queues, it's not required to be `Sync` on the
+/// consumer side. Thread pool will use one consumer local queue per thread,
+/// which make it possible to do extreme optimizations and define complicated
+/// data struct.
+pub trait TaskQueue: Clone {
+    type Consumer;
+    type Task;
+
+    /// Creates a queue with a promise to only use at most `con` consumers
+    /// at the same time.
+    fn with_consumers(con: usize) -> Self;
+
+    /// Pushes a task to the queue.
+    ///
+    /// If the queue is closed, the method should behave like no-op.
+    fn push(&self, task: Self::Task);
+
+    /// Closes the queue so that no more tasks can be scheduled.
+    fn close(&self);
+}
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -59,9 +59,12 @@ pub trait Runner {
 
 /// Allows spawning a task to the thread pool from a different thread.
 pub trait RemoteSpawn: Sync + Send {
+    /// The task it can spawn.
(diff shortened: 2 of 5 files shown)


src/runner.rs
```diff
@@ -79,18 +82,23 @@ impl<S: RemoteSpawn> RemoteSpawnExt for S {}
 
 /// Allows spawning a task inside the thread pool.
 pub trait LocalSpawn {
+    /// The task it can spawn.
     type Task;
+    /// The context that associated with the task.
```
sticnarf, line 87:
```suggestion
    /// The context associated with the task.
```

src/queue.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+/// A Task queue for thread pool.
+///
+/// Unlike a general MPMC queues, it's not required to be `Sync` on the
+/// consumer side. Thread pool will use one consumer local queue per thread,
+/// which make it possible to do extreme optimizations and define complicated
+/// data struct.
+pub trait TaskQueue: Clone {
```
sticnarf, line 9: I'm a bit puzzled about what a `TaskQueue` is. Judging by its name and doc, I would consider it like an MPMC queue. However, it seems to me that it is also responsible for managing the threads, like the `QueueCore` in the original code.

I'm thinking if it's better to leave thread-related work to the thread pool itself and make the queue just like an MPMC queue.
My thought is making `Remote` able to not only push tasks to the queue but also unpark a thread. The queue here provides a method for the thread pool to create specified number of consumers. The thread pool can use the consumers to create worker threads.

src/queue.rs
```diff
@@ -0,0 +1,24 @@
...
+    type Task;
+
+    /// Creates a queue with a promise to only use at most `con` consumers
+    /// at the same time.
+    fn with_consumers(con: usize) -> Self;
+
+    /// Pushes a task to the queue.
+    ///
+    /// If the queue is closed, the method should behave like no-op.
+    fn push(&self, task: Self::Task);
```
sticnarf, line 20: The main goal of this thread pool is to adjust the number of running threads smartly. So I think we may force the element in the queue to be something like the `SchedUnit` in the original code, otherwise we cannot know the spawn time. Additionally, I think `SpawnOption` can be also added to `SchedUnit` and `Remote::spawn` may accept a `SpawnOption` as well.

src/queue.rs
BusyJay: That's exact what I'm planning to do.

sticnarf: Seems that the `with_consumers` method is not enough. There need be another method to fetch the created consumers.

BusyJay: The trait is not completed, just listed necessary methods to provide thread pool interfaces.

sticnarf (approved):
Let's move forward. Some changes can be done in the future.

PR #5, opened 2019-11-27
use TaskCell everywhere and remove TaskContext

sticnarf:
This PR adds `TaskCell` trait. With this trait, the queue can get the spawn options from it. To spawn a task, one should use spawn a `TaskCell`.

`TaskContext` is removed. The only use of `TaskContext` is for spawning through `LocalSpawn`. The runner actually cannot utilize any data in it. So it can be managed by the queue itself and needn't be known by the runner.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -8,7 +8,7 @@
 /// data struct.
 pub trait TaskQueue: Clone {
     type Consumer;
-    type Task;
+    type TaskCell;
 
     /// Creates a queue with a promise to only use at most `con` consumers
     /// at the same time.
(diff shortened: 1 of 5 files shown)


src/queue.rs
```diff
@@ -1,5 +1,19 @@
...
+    /// Extra information in the cell.
+    type Extras;
+
+    /// Gets mutable extra information.
+    ///
+    /// # Safety
+    ///
+    /// This method is not safe. The caller must ensure that he is the only
+    /// one accessing the extras.
+    unsafe fn mut_extras(&mut self) -> &mut Self::Extras;
```
BusyJay: I think it should be `&self` instead. Rust guarantees only one can access a `&mut` method.

BusyJay (approved):

breezewish (approved):

PR #6, opened 2019-11-28
implement std Future runner

sticnarf:
This PR implements the runner for `std::future::Future`.


src/task/future.rs
```diff
@@ -0,0 +1,405 @@
...
+use std::mem::ManuallyDrop;
+use std::pin::Pin;
+use std::sync::atomic::{AtomicU8, Ordering::SeqCst};
+use std::sync::Arc;
+use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
+use std::{fmt, mem};
+
+/// A [`Future`] task.
+pub struct Task<Remote, Extras> {
+    status: AtomicU8,
```
AndreMouche, line 23: How about defining an enum for the status?

src/task/future.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+}
+
+#[inline]
+unsafe fn wake_ref_raw<Remote, Extras>(this: *const ())
+where
+    Remote: RemoteSpawn<TaskCell = TaskCell<Remote, Extras>>,
+{
+    let task = ManuallyDrop::new(task_cell(this as *const Task<Remote, Extras>).0);
+    let mut status = task.status.load(SeqCst);
```
AndreMouche: It seems logical for L149-176 is the same as L114-140, could we reuse one?

src/task/future.rs
sticnarf: The status needs to support atomic operations. We cannot switch between enum variants atomically.

nrc:
Now that #5 is merged, could you rebase this PR to make it easier to review please?

sticnarf:
@nrc Sorry, this PR originally has different history from master... I've cleaned up the commit history...

nrc (dismissed):
LGTM!

src/task/future.rs
```diff
@@ -0,0 +1,387 @@
...
+    status: AtomicU8,
+    future: UnsafeCell<Pin<Box<dyn Future<Output = ()> + Send + 'static>>>,
+    remote: Remote,
+    extras: UnsafeCell<Extras>,
+}
+
+/// A [`Future`] task cell.
+pub struct TaskCell<Remote, Extras>(Arc<Task<Remote, Extras>>);
+
+unsafe impl<Remote: Send + Sync, Extras> Send for TaskCell<Remote, Extras> {}
```
BusyJay: I think `Extras` needs to be `Send`.

sticnarf: Indeed.

src/task/future.rs
```diff
@@ -0,0 +1,388 @@
...
+    remote: Remote,
+    extras: UnsafeCell<Extras>,
+}
+
+/// A [`Future`] task cell.
+pub struct TaskCell<Remote, Extras>(Arc<Task<Remote, Extras>>);
+
+// Safety: It is ensured that `future` and `extras` are always accessed by
+// only one thread at the same time.
+unsafe impl<Remote: Sync, Extras> Sync for Task<Remote, Extras> {}
```
sticnarf, line 34: I change the original code to `unsafe impl` for `Task`.
PTAL the safety insurance @BusyJay @nrc

BusyJay: I think `Extras` needs to be `Send` to make `TaskCell::mut_extras` safe, otherwise it will allow to send `Extras` to other thread no matter `Extras` is `Send` or not.

sticnarf: I don't get it. `TaskCell` is `Send` and `Sync` only if `Extras` is `Send`, so `extras` won't be sent to other threads.

BusyJay: Never mind, seems `TaskCell::mut_extras` takes a mut reference, it should be safe.

nrc (dismissed):

src/task/future.rs
```diff
@@ -0,0 +1,388 @@
...
+    Remote: RemoteSpawn<TaskCell = TaskCell<Remote, Extras>>,
+{
+    let task = &task_cell.0;
+    let mut status = task.status.load(SeqCst);
+    loop {
+        match status {
+            WAITING => {
+                match task
+                    .status
+                    .compare_exchange(WAITING, POLLING, SeqCst, SeqCst)
```
BusyJay: I find it quite confusing that marking woken up task as `POLLING`, because it actually has not been polled yet.

I suggest to use following state instead. 
```
const NOTIFIED: u8 = 1;
const IDLE: u8 = 2;
const POLLING: u8 = 3;
const COMPLETED: u8 = 4;
```

Ref: https://github.com/tikv/grpc-rs/blob/0c16249ca5afabbbf308e9d6dda9804ffcf0535a/src/task/executor.rs#L73-L84

src/task/future.rs
```diff
@@ -0,0 +1,388 @@
...
+                        task.remote.spawn(clone_task(&**task));
+                        break;
+                    }
+                    Err(cur) => status = cur,
+                }
+            }
+            POLLING => {
+                match task
+                    .status
+                    .compare_exchange(POLLING, REPOLL, SeqCst, SeqCst)
```
BusyJay: Since it's in a loop, I think you can use `compare_exchange_weak`.

src/task/future.rs
sticnarf: Updated. PTAL again.

src/task/future.rs
```diff
@@ -0,0 +1,393 @@
...
+    R: RemoteSpawn<TaskCell = TaskCell<R, Extras>>,
+{
+    type Spawn = L;
+
+    fn handle(&mut self, _local: &mut L, task_cell: TaskCell<R, Extras>) -> bool {
+        let task = task_cell.0;
+        unsafe {
+            let waker = ManuallyDrop::new(waker(&*task));
+            let mut cx = Context::from_waker(&waker);
+            loop {
```
BusyJay, line 227: I think there should be a limit for the loop, otherwise it may starve other futures.

src/task/future.rs
```diff
@@ -0,0 +1,393 @@
...
+                task.status.store(POLLING, SeqCst);
+                if let Poll::Ready(_) = Pin::new(&mut *task.future.get()).poll(&mut cx) {
```
BusyJay: Why not `(&mut *task.future.get()).as_mut().poll(&mut cx)`?

src/task/future.rs
```diff
@@ -0,0 +1,393 @@
...
+                    task.status.store(COMPLETED, SeqCst);
+                    return true;
+                }
+                match task.status.compare_exchange(POLLING, IDLE, SeqCst, SeqCst) {
+                    Ok(_) => return false,
+                    Err(cur) if cur == NOTIFIED => continue,
```
BusyJay: You mean `Err(NOTIFIED)`?

src/task/future.rs
```diff
@@ -0,0 +1,393 @@
...
+
+#[inline]
+unsafe fn drop_raw<Remote, Extras>(this: *const ())
+where
+    Remote: RemoteSpawn<TaskCell = TaskCell<Remote, Extras>>,
+{
+    drop(task_cell(this as *const Task<Remote, Extras>))
+}
+
+unsafe fn wake_impl<Remote, Extras>(task_cell: &TaskCell<Remote, Extras>)
```
BusyJay, line 120: How about taking `Cow<Arc<Task>>`? So that it can save one clone for `wake_up`.

src/task/future.rs
```diff
@@ -0,0 +1,393 @@
...
+            res_tx.send(2).unwrap();
+        };
+        local.spawn(TaskCell::new(fut, local.remote(), ()));
+
+        local.handle_once();
+        assert_eq!(res_rx.recv().unwrap(), 1);
+        assert!(res_rx.try_recv().is_err());
+
+        let waker = waker_rx.recv().unwrap();
+        f(waker);
```
BusyJay, line 362: How to verify it's really woken?

src/task/future.rs
sticnarf: I thought `Err(NOTIFIED)` would make it bind value to a variable called `NOTIFIED`...
Seems I worried too much.

BusyJay: Only when `NOTIFIED` is not a const. Although it's a confusing part of rust.

src/task/future.rs
sticnarf: I think there's no difference. I'll change if you prefer it.

BusyJay: It's confusing to create a new `Pin` when the inner type is `Pin` already.

src/task/future.rs
sticnarf: I verify it by the code below. If `2` is received, the future must have been woken. Do you think the verification is too implicit? Would it be better to record it's spawned in `MockRemote` and check?

BusyJay: Get it.

src/task/future.rs
sticnarf: We can't. We need `remote` in the task to spawn. So even if we have the ownership of the task, we cannot spawn it with `remote` still in use. That means we need either cloning remote or task, which is both an atomic operation.

BusyJay (approved):

AndreMouche:
Please fix the conflicts

AndreMouche (approved):
LGTM

PR #7, opened 2019-12-02
complete task queue traits

sticnarf:
* Remove the trait function creating the queue. Instead, use free functions to create the pair of `TaskInjector` and `LocalQueue`s.
* Add `LocalQueue` trait.
* `LocalQueue::pop` returns the instant when the task was pushed to the queue and whether the task is from the current local queue.


src/queue/mod.rs
```diff
@@ -16,18 +20,25 @@ pub trait TaskCell {
...
-    /// If the queue is closed, the method should behave like no-op.
     fn push(&self, task: Self::TaskCell);
+}
+
+/// The consumer of a task queue.
+pub trait Consumer {
+    /// The task cell in the queue.
+    type TaskCell: TaskCell;
 
-    /// Closes the queue so that no more tasks can be scheduled.
```
BusyJay: Why remove it?

sticnarf: What's the purpose of it?

BusyJay: As the comment says, close the queue so that no more tasks can be scheduled. See also crossbeam-rs/crossbeam#236.

sticnarf: Should it be the thread pool who prevents the tasks from being schduled?

BusyJay: How can it prevent scheduling by `Remote`?

BusyJay: What I thought is to reduce an atomic load by handling it inside the queue. But if level queue can't benefit from it, I also feel OK to do it in the pool part.

sticnarf: I'm curious how it reduces an atomic load? The closed flag  should be shared among all senders. Change the inner code of `crossbeam_deque::Injector::push`?

BusyJay: It depends on the implement details. For the original queue implement, the desired way is to change the code of `Injector`.

src/queue/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// The local queue of a task queue.
+pub trait LocalQueue {
+    /// The task cell in the queue.
+    type TaskCell: TaskCell;
+
+    /// Pushes a task to the local queue.
+    fn push(&mut self, task_cell: Self::TaskCell);
+
+    /// Gets a task cell from the queue. Rturns `None` if there is no task cell
```
BusyJay: s/Rturns/Returns/

BusyJay (commented):
Rest LGTM

src/queue/mod.rs
sticnarf: Fixed.

BusyJay (approved):

AndreMouche (approved):
LGTM

PR #8, opened 2019-12-02
implement simple queue

### LINKED ISSUE #10: implement simple task queue
A simple task queue supports work stealing and uses the First Come First Serve (FCFS) strategy. Tasks pushed to the injector first are popped from the injector first. However, the local queue is a LIFO queue for efficiency.

Just use a single `crossbeam_deque::Injector`as the global queue and several `crossbeam_deque::Worker`s as the local queues.

sticnarf:
Based on #7 .

This PR implements a simple task queue with crossbeam deque. It has only one injector and supports stealing tasks from other workers.

Fixes #10


src/queue/simple.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+        let mut need_retry;
+        loop {
+            need_retry = false;
+            match self.injector.steal_batch_and_pop(&self.local_queue) {
+                Steal::Success(t) => return Some(into_pop(t, false)),
+                Steal::Retry => need_retry = true,
+                _ => {}
+            }
+            for (i, stealer) in self.stealers.iter().enumerate() {
```
BusyJay: Technically it can lead to unbalance because tasks of queues at smaller index will always get more chance to be stolen.

sticnarf: OK. I can use a rng to solve it.

BusyJay (dismissed):

BusyJay (approved):

AndreMouche (approved):
LGTM

PR #14, opened 2019-12-05
get rid of "mod.rs"s

BusyJay:
So we don't have to distinguish different mod files.


sticnarf (approved):
LGTM. I didn't realize Rust 2018 prefers the new style.

AndreMouche (approved):
LGTM

PR #15, opened 2019-12-06
implement multilevel task queue

sticnarf:
This PR implements a multilevel task queue. 

Tasks are pushed to the queue  of corresponding level according to the running time of the task. Tasks in each level have different chances of running.


src/queue/multilevel.rs
```diff
@@ -0,0 +1,885 @@
...
+
+/// The extras for the task cells pushed into a multilevel task queue.
+#[derive(Debug, Clone)]
+pub struct MultilevelQueueExtras {
+    /// The identifier of the task.
+    task_id: u64,
+    /// The instant when the task cell is pushed to the queue.
+    schedule_time: Option<Instant>,
+    /// The time spent on handling this task.
+    running_time: Option<Arc<ElapsedTime>>,
```
BusyJay: Why use `Option`? Is `ElapsedTime::zero()` sufficient enough?

src/queue/multilevel.rs
```diff
@@ -0,0 +1,885 @@
...
+    }
+}
+
+/// The extras for the task cells pushed into a multilevel task queue.
+#[derive(Debug, Clone)]
+pub struct MultilevelQueueExtras {
+    /// The identifier of the task.
+    task_id: u64,
+    /// The instant when the task cell is pushed to the queue.
+    schedule_time: Option<Instant>,
```
BusyJay: Maybe `TimeSpec` is a more suitable type as we don't require precise clock.

src/queue/multilevel.rs
sticnarf: `running_time` is created and returned by `TaskElapsedMap`. So we don't need the extra allocation caused by `Arc` when initializing `MultilevelQueueExtras`.

src/queue/extras.rs
```diff
@@ -1,12 +1,58 @@
...
+            current_level: 0,
+            fixed_level: None,
+        }
+    }
+
+    /// Creates a default `Extras` for task cells pushed into a multilevel task
+    /// queue. It generates a random task id and does not specify the fixed
+    /// level.
+    pub fn multilevel_default() -> Extras {
+        Self::new_multilevel(thread_rng().next_u64(), None)
```
BusyJay, line 44: Why not use UUID or a default atomic counter?

sticnarf: Task ID can be passed from the client so an atomic counter seems weird in semantics.

I will change it to UUID. I just thought 64 bits is enough. According to birthday attack, if we store 1e7 random u64 ids, the possibility of id collision is only 2.7e-6.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,852 @@
...
+/// The local queue of a multilevel task queue.
+pub struct QueueLocal<T> {
+    local_queue: Worker<T>,
+    level_injectors: Arc<[Injector<T>; LEVEL_NUM]>,
+    stealers: Vec<Stealer<T>>,
+    manager: Arc<LevelManager>,
+    rng: SmallRng,
+}
+
+impl<T> QueueLocal<T> {}
```
BusyJay: What's the purpose?

src/queue/multilevel.rs
```diff
@@ -0,0 +1,852 @@
...
+            manager: self.manager.clone(),
+        }
+    }
+}
+
+/// The runner for multilevel task queues.
+///
+/// The runner helps multilevel task queues collect additional information.
+/// [`MultilevelRunnerBuilder`] is the [`RunnerBuilder`] for this runner.
+pub struct MultilevelRunner<R> {
```
BusyJay, line 184: Better move to `src/task/`.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,859 @@
...
+        )
+    }
+}
+
+thread_local!(static RECENT_NOW: Cell<Instant> = Cell::new(Instant::now()));
+
+/// Returns an instant corresponding to now and updates the thread-local recent
+/// now.
+///
+/// You should only use it when the thread-local recent now is recently updated.
```
BusyJay: Shouldn't it be only used when recent now is *not* recently updated?

src/queue/multilevel.rs
```diff
@@ -0,0 +1,859 @@
...
+        // The statistics may be not so accurate because we cannot load two
+        // atomics at the same time.
+        let total = self.total_elapsed.0.load(SeqCst);
+        if Duration::from_micros(total) < ADJUST_CHANCE_INTERVAL {
+            // Another thread just adjusted the chances.
+            return;
+        }
+        let total = self.total_elapsed.0.swap(0, SeqCst);
+        let level0 = self.level0_elapsed.0.swap(0, SeqCst);
+        let current_proportion = level0 as f64 / total as f64;
```
BusyJay, line 286: What if total is zero?

sticnarf: `total` must be at least `ADJUST_CHANCE_INTERVAL` thus cannot be zero.

BusyJay: But it just gets swapped at L300.

sticnarf: `maybe_adjust_chance` is locked by `adjusting`, so there is only one thread executing this function. `total_elapsed` increases monotonously except the swap at L300. It means the total at L300 is not less than the total at L295 so it cannot be zero.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,859 @@
...
+{
+    pub(super) fn push(&self, mut task_cell: T) {
+        self.manager.prepare_before_push(&mut task_cell);
+        let level = task_cell.mut_extras().current_level as usize;
+        self.level_injectors[level].push(task_cell);
+    }
+}
+
+/// The local queue of a multilevel task queue.
+pub struct QueueLocal<T> {
```
BusyJay, line 71: `QueueLocal` and `LocalQueue` are a bit confusing. Maybe we can unify the naming in another PR.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,859 @@
...
+    pub fn runner_builder<B>(&self, inner_runner_builder: B) -> MultilevelRunnerBuilder<B> {
+        MultilevelRunnerBuilder {
+            inner: inner_runner_builder,
+            manager: self.manager.clone(),
+        }
+    }
+
+    fn build_raw<T>(self, local_num: usize) -> (QueueInjector<T>, Vec<QueueLocal<T>>) {
+        let level_injectors: Arc<[Injector<T>; LEVEL_NUM]> =
+            Arc::new(InitWith::init_with(Injector::new));
```
BusyJay: Seems unnecessary to introduce an extra dependency.

Cargo.toml
```diff
@@ -12,3 +12,6 @@ repository = "https://github.com/tikv/yatp/"
 num_cpus = "1.0"
 crossbeam-deque = "0.7"
 rand = { version = "0.7", features = ["small_rng"] }
+init_with = "1.1.0"
+dashmap = "1.2.0"
```
BusyJay, line 15: 2.0.x is available now.

src/queue/multilevel.rs
sticnarf: Oh, it's a mistake. It should belong to the `recent` function.

sticnarf:
I'll address the comments after #18 is merged.

Cargo.toml
sticnarf: 2.x seems not stable yet and lacks tests. It mightn't be a good idea to upgrade.

src/queue/multilevel.rs
sticnarf: The runner is queue specific. It is not a real kind of task. I think it a bit weird to put it in `src/task/` alongwith future and callback.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,846 @@
...
+use rand::prelude::*;
+use std::cell::Cell;
+use std::cmp;
+use std::iter;
+use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, AtomicUsize, Ordering::SeqCst};
+use std::sync::{Arc, RwLock};
+use std::time::{Duration, Instant};
+use uuid::Uuid;
+
+const LEVEL_NUM: usize = 3;
```
BusyJay, line 22: Is 3 an appropriate level number for most cases?

src/queue/multilevel.rs
```diff
@@ -0,0 +1,838 @@
...
+    }
+}
+
+thread_local!(static TLS_LAST_CLEANUP_TIME: Cell<Instant> = Cell::new(Instant::now()));
+
+struct TaskElapsedMap {
+    new_index: AtomicUsize,
+    maps: [DashMap<u64, Arc<ElapsedTime>>; 2],
+    cleanup_interval: Duration,
+    last_cleanup_time: RwLock<Instant>,
```
BusyJay: `Mutex` is better.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,838 @@
...
+    }
+
+    fn get_elapsed(&self, key: u64) -> Arc<ElapsedTime> {
+        let new_index = self.new_index.load(SeqCst);
+        let new_map = &self.maps[new_index];
+        let old_map = &self.maps[new_index ^ 1];
+        if let Some(v) = new_map.get(&key) {
+            return v.clone();
+        }
+        let elapsed = match old_map.remove(&key) {
```
BusyJay: I think there is a race that two tasks with the same key arrive and overwrite the original elapsed with a default one.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,838 @@
...
+        let res = self.inner.handle(local, task_cell);
+        let elapsed = begin.elapsed();
+        if let Some(running_time) = running_time {
+            running_time.inc_by(elapsed);
+        }
+        if level == 0 {
+            self.manager.level0_elapsed.inc_by(elapsed);
+        }
+        let current_total = Duration::from_micros(
+            self.manager.total_elapsed.inc_by(elapsed) + elapsed.as_micros() as u64,
```
BusyJay: Why add twice?

sticnarf: `fetch_add` returns the previous value.

src/queue/multilevel.rs
sticnarf: Why is `Mutex` better? Multiple threads may read the value but only one will modify it. Using a `RwLock` enables multiple threads to read the value at the same time.

src/queue/multilevel.rs
sticnarf: Having many levels harms the performance while having few levels makes scheduling too rough. I just think 3 can be a good choice but haven't tried other numbers.

src/queue/multilevel.rs
sticnarf: Say T1 removes elapsed from old first, then T2 cannot find elapsed from old.
T1 will `insert` the elapsed into new, T2 will `get_or_insert_with` default from new.
If T1 `insert` first, T2 will get the inserted one.
If T2 `get_or_insert_with` first, T2 gets a default elapsed. But the one in new map will be soon overwritten by T1.
So the problem is T2 gets the wrong one.
I change it to not removing from old.

src/queue/multilevel.rs
BusyJay: Because `Mutex` is faster than `RwLock`, unless read lock is held for a long time, `Mutex` should always have the best performance.

src/queue/multilevel.rs
BusyJay: It's a bit confusing. I think we should make `inc_by` return the current value.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,835 @@
...
+        if let Some(v) = new_map.get(&key) {
+            return v.clone();
+        }
+        let elapsed = match old_map.get(&key) {
+            Some(v) => {
+                new_map.insert(key, v.clone());
+                v.clone()
+            }
+            _ => {
+                let v = new_map.get_or_insert_with(&key, Default::default);
```
BusyJay, line 363: What if `new_index` is changed during two concurrent `get_elapsed`?

sticnarf: The two elapsed can be different. It can make the statistics inaccurate but I can't find out a good way to solve it now.

src/queue/multilevel.rs
```diff
@@ -0,0 +1,834 @@
...
+/// The runner for multilevel task queues.
+///
+/// The runner helps multilevel task queues collect additional information.
+/// [`MultilevelRunnerBuilder`] is the [`RunnerBuilder`] for this runner.
+pub struct MultilevelRunner<R> {
+    inner: R,
+    manager: Arc<LevelManager>,
+}
+
+impl<R, T> Runner for MultilevelRunner<R>
```
BusyJay, line 189: Should be `MultiLevelRunner`.

BusyJay (commented):
Rest LGTM

src/queue/multilevel.rs
sticnarf: Here I suppose "multilevel" is a single word. If you prefer "multi-level", we should change all occurances to "MultiLevel" (for camel case) or "multi_level" (for snake case).

BusyJay: Oh, I didn't realize it was a word.

BusyJay (approved):

AndreMouche (approved):
LGTM

PR #16, opened 2019-12-18
queue: use enum instead of generic

BusyJay:
Due to limitation of rust-lang/rust#35978, using generic requires trait
object or redundant verbose struct wrapper to work. So this PR uses enum
instead to get around the problem and make the types simpler.

### DIFF
--- a/src/queue/extras.rs
+++ b/src/queue/extras.rs
@@ -4,8 +4,7 @@ use std::time::Instant;
 
 /// The extras for the task cells pushed into a simple task queue.
 ///
-/// [`Default::default`] can be used to create a [`SimpleQueueExtras`] for
-/// a task cell.
+/// [`Default::default`] can be used to create a [`Extras`] for a task cell.
 #[derive(Debug, Default, Clone, Copy)]
 pub struct Extras {
     /// The instant when the task cell is pushed to the queue.
--- a/src/queue/extras.rs
+++ b/src/queue/extras.rs
@@ -1,0 +1,13 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::time::Instant;
+
+/// The extras for the task cells pushed into a simple task queue.
+///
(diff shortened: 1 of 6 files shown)


src/queue/extras.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::time::Instant;
+
+/// The extras for the task cells pushed into a simple task queue.
+///
+/// [`Default::default`] can be used to create a [`SimpleQueueExtras`] for
+/// a task cell.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Extras {
```
sticnarf, line 9: How to add new types of extras when we want to add a new queue type?

BusyJay: Just add new fields. `Extras` should be small enough to be shared.

sticnarf: https://github.com/tikv/yatp/pull/15/files#diff-15709e727979b90faec03201bc3303baR63
Current extras for leveled queue has quite a few fields and is not so small... And it may be confusing to mix different extras.

BusyJay: Yes, I have looked into it. It add extra 24 bytes, which is the same size of `Option<Instant>`.

src/queue/extras.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
+
+use std::time::Instant;
+
+/// The extras for the task cells pushed into a simple task queue.
+///
+/// [`Default::default`] can be used to create a [`SimpleQueueExtras`] for
+/// a task cell.
```
sticnarf: These comments need updating.

sticnarf (commented):
Rest LGTM

sticnarf (approved):

AndreMouche (approved):
LGTM

PR #17, opened 2019-12-20
pool: adjust layout

BusyJay:
This PR splits pool into several submodels. `Runner` is also moved into
pool as it's an implement details of the pool, should not be exposed to
users in most cases.

### DIFF
--- a/src/pool/runner.rs
+++ b/src/pool/runner.rs
@@ -1,5 +1,7 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
+use crate::pool::spawn::LocalSpawn;
+
 /// In the model of yatp, any piece of logic aiming to be executed in a thread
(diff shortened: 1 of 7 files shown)


src/pool/runner.rs
```diff
@@ -84,3 +63,14 @@ pub trait RunnerBuilder {
     /// Builds a runner.
     fn build(&mut self) -> Self::Runner;
 }
+
+/// A builder that create new Runner by cloning the old one.
+pub struct CloneRunnerBuilder<R>(pub R);
```
sticnarf, line 68: What's the purpose of this builder?

BusyJay: Provides a convenient `Builder` so that users don't have to write their own. Cloning is a very common use case especially for tests.

sticnarf (approved):
LGTM

AndreMouche (approved):
LGTM

PR #18, opened 2019-12-20
*: add basic pool implement

BusyJay:
This PR adds the basic pool implement, including task spawning and thread waking up mechanism.

### DIFF
--- a/src/pool/runner.rs
+++ b/src/pool/runner.rs
@@ -28,27 +28,27 @@ pub trait Runner {
     /// Called when the runner is started.
     ///
     /// It's guaranteed to be the first method to call before anything else.
-    fn start(&mut self, _spawn: &mut Local<Self::TaskCell>) {}
+    fn start(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
     /// Called when a task needs to be handled.
     ///
     /// It's possible that a task can't be finished in a single execution, in
     /// which case feel free to spawn the task again and return false to
     /// indicate the task has not been finished yet.
-    fn handle(&mut self, spawn: &mut Local<Self::TaskCell>, task_cell: Self::TaskCell) -> bool;
+    fn handle(&mut self, local: &mut Local<Self::TaskCell>, task_cell: Self::TaskCell) -> bool;
 
     /// Called when the runner is put to sleep.
-    fn pause(&mut self, _spawn: &mut Local<Self::TaskCell>) -> bool {
+    fn pause(&mut self, _local: &mut Local<Self::TaskCell>) -> bool {
         true
     }
 
     /// Called when the runner is woken up.
-    fn resume(&mut self, _spawn: &mut Local<Self::TaskCell>) {}
+    fn resume(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
     /// Called when the runner is about to be destroyed.
     ///
     /// It's guaranteed that no other method will be called after this method.
-    fn end(&mut self, _spawn: &mut Local<Self::TaskCell>) {}
+    fn end(&mut self, _local: &mut Local<Self::TaskCell>) {}
 }
 
 /// A builder trait that produce `Runner`.
--- a/src/pool/runner.rs
+++ b/src/pool/runner.rs
@@ -64,6 +64,7 @@ pub trait RunnerBuilder {
     fn build(&mut self) -> Self::Runner;
 }
 
+/// A builder that create new Runner by cloning the old one.
 pub struct CloneRunnerBuilder<R>(pub R);
 
 impl<R: Runner + Clone> RunnerBuilder for CloneRunnerBuilder<R> {
--- a/src/pool/runner.rs
+++ b/src/pool/runner.rs
@@ -35,11 +35,7 @@ pub trait Runner {
     /// It's possible that a task can't be finished in a single execution, in
     /// which case feel free to spawn the task again and return false to
     /// indicate the task has not been finished yet.
-    fn handle(
-        &mut self,
-        spawn: &mut Local<Self::TaskCell>,
-        task_cell: Self::TaskCell,
-    ) -> bool;
+    fn handle(&mut self, spawn: &mut Local<Self::TaskCell>, task_cell: Self::TaskCell) -> bool;
 
     /// Called when the runner is put to sleep.
     fn pause(&mut self, _spawn: &mut Local<Self::TaskCell>) -> bool {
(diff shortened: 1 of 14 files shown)


src/pool/runner.rs
```diff
@@ -23,36 +23,32 @@ use crate::pool::spawn::LocalSpawn;
...
 pub trait Runner {
     /// The local spawn that can be accepted to spawn tasks.
-    type Spawn: LocalSpawn;
+    type TaskCell;
 
     /// Called when the runner is started.
     ///
     /// It's guaranteed to be the first method to call before anything else.
-    fn start(&mut self, _spawn: &mut Self::Spawn) {}
+    fn start(&mut self, _spawn: &mut Local<Self::TaskCell>) {}
```
sticnarf: Maybe the name `spawn` should be changed after `LocalSpawn` is changed to a concrete type `Local`.

src/pool/spawn.rs
```diff
@@ -1,56 +1,276 @@
...
+        &self.core
+    }
+
+    pub(crate) fn pop(&mut self) -> Option<Pop<T>> {
+        self.local_queue.pop()
+    }
+
+    /// Pops a task from the queue.
+    ///
+    /// If there is no tasks at the moment, it will go to sleep until woken
```
sticnarf: `is` -> `are`

src/pool/worker.rs
```diff
@@ -0,0 +1,42 @@
...
+    T: TaskCell + Send,
+    R: Runner<TaskCell = T>,
+{
+    pub fn run(mut self) {
+        self.runner.start(&mut self.local);
+        while !self.local.core().is_shutdown() {
+            let task = match self.local.pop() {
+                Some(t) => t,
+                None => {
+                    self.runner.pause(&mut self.local);
```
sticnarf, line 28: The thread may fail to park because a new task is added between the previous `pop` and the `pop` in park's validate function. Should this be moved to park's `before_sleep`?

src/pool/worker.rs
```diff
@@ -0,0 +1,42 @@
...
+                    match self.local.pop_or_sleep() {
+                        Some(t) => {
+                            self.runner.resume(&mut self.local);
```
sticnarf, line 31: And to make the behavior more accurate, the resume method should be called when the thread is really unparked.

src/pool/spawn.rs
```diff
@@ -1,56 +1,276 @@
...
+    /// up by other threads.
+    pub(crate) fn pop_or_sleep(&mut self) -> Option<Pop<T>> {
+        let address = &*self.core as *const QueueCore<T> as usize;
+        let mut task = None;
+        let mut timeout = Some(Instant::now() + self.core.config.max_idle_time);
```
sticnarf: I can't understand what the idle time means.
Now it seems to be the maximum time the thread waits for the first time and try to pop again. And at the second time, it waits forever until unparked.

src/pool/spawn.rs
```diff
@@ -1,56 +1,276 @@
...
+            };
         }
     }
 }
+
+/// Building remotes and locals from the given queue and configuration.
+///
+/// This is only for tests purpose so that a thread pool doesn't have to be
+/// spawned to test a Runner.
+#[doc(hidden)]
```
sticnarf: It seems a bit rough to just hide it from docs. How about placing it in a `test_util` mod?

sticnarf: Actually I think an ideal unit test of a runner should not depend on `QueueCore`, which is another rather complex component.

src/pool/spawn.rs
```diff
@@ -1,56 +1,276 @@
...
+        QueueCore {
+            global_queue,
+            active_workers: AtomicUsize::new(config.max_thread_count << WORKER_COUNT_SHIFT),
+            config,
+        }
+    }
+
+    /// Ensures there are enough workers to handle pending tasks.
+    ///
+    /// If the method is going to wake up any threads, source is used to trace who triggers
```
sticnarf, line 52: What's the purpose of tracing who triggers the action?

src/pool/spawn.rs
```diff
@@ -1,56 +1,276 @@
...
+        let address = &*self.core as *const QueueCore<T> as usize;
+        let mut task = None;
+        let mut timeout = Some(Instant::now() + self.core.config.max_idle_time);
+        let id = self.id;
+        loop {
+            let res = unsafe {
+                parking_lot_core::park(
+                    address,
+                    || {
+                        if timeout.is_some() && !self.core.mark_sleep() || self.core.is_shutdown() {
```
sticnarf: Will it reduce atomic operations on if we move `mark_sleep` to `before_sleep`?

src/pool/spawn.rs
BusyJay: For debugging and performance analysis.

src/pool/spawn.rs
BusyJay: It's wrong to `mark_sleep` in `before_sleep`. It requires lock to avoid race between "push", "wake" and "sleep".

src/pool/spawn.rs
BusyJay: > ...an ideal unit test of a runner ...

Due to the removal of `Spawn` trait, there has to be a way to create a `Local` type.

> ... is another rather complex component.

Indeed. So the function is provided to hide the complexity.

src/pool/worker.rs
BusyJay: Maybe we should just update the comment to indicate the runner is *likely* to go to sleep. Not to put it to `before_sleep` to be better modularization.

BusyJay:
Very strange that all tests pass on local mac machine. 🤔

sticnarf:
I have no idea why it fails as well...

sticnarf (approved):
LGTM

src/pool/spawn.rs
```diff
@@ -1,56 +1,269 @@
...
 
-    /// Spawns a task into the thread pool.
-    fn spawn(&mut self, task: Self::TaskCell);
+    /// Ensures there are enough workers to handle pending tasks.
+    ///
+    /// If the method is going to wake up any threads, source is used to trace who triggers
+    /// the action.
+    pub fn ensure_workers(&self, source: usize) {
+        let cnt = self.active_workers.load(Ordering::Relaxed);
+        if (cnt >> WORKER_COUNT_SHIFT) >= self.config.max_thread_count || is_shutdown(cnt) {
```
AndreMouche, line 56: how about defining an independent function to compute the thread count? which is similar with is_shutdown

AndreMouche (approved):
LGTM

PR #19, opened 2019-12-25
pool: fix ordering issue

BusyJay:
loom shows that load and store can be reordered when using relax order,
so change it to SeqCst instead.

### DIFF
--- a/src/queue/simple.rs
+++ b/src/queue/simple.rs
@@ -46,7 +46,6 @@ pub struct QueueLocal<T> {
     local_queue: Worker<T>,
     injector: Arc<Injector<T>>,
     stealers: Vec<Stealer<T>>,
-    rng: SmallRng,
 }
 
 impl<T> QueueLocal<T>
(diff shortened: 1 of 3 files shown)


src/queue/simple.rs
```diff
@@ -74,36 +73,34 @@ where
...
-            if len > 0 {
-                let start_index = self.rng.gen_range(0, len);
-                for stealer in self
-                    .stealers
-                    .iter()
-                    .chain(&self.stealers)
-                    .skip(start_index)
-                    .take(len)
-                {
+            if !self.stealers.is_empty() {
```
sticnarf, line 84: This line seems unnecessary

BusyJay: It reduces several instructions, but I'm OK to remove it if you insist as it does not have significant performance impact in most case.

sticnarf (dismissed):

src/queue/simple.rs
```diff
@@ -46,7 +46,6 @@ pub struct QueueLocal<T> {
     local_queue: Worker<T>,
     injector: Arc<Injector<T>>,
     stealers: Vec<Stealer<T>>,
-    rng: SmallRng,
```
sticnarf, line 49: the small_rng feature can be removed from `Cargo.toml`

sticnarf (approved):

AndreMouche (approved):
LGTM

PR #20, opened 2019-12-26
support proactive rescheduling for futures

sticnarf:
If a future is waked on polling, it is repolled immediately. So it is impossible for a user to give up timeslice proactively to the scheduler.

This PR adds an async function to reschedule the current future task.


BusyJay (dismissed):
I prefer setting `repoll_limit` to 1 as `reschedule` does not work for other executors. But I'm OK with the change.

AndreMouche (dismissed):
LGTM

sticnarf:
PTAL again @AndreMouche @BusyJay

BusyJay (approved):

PR #21, opened 2019-12-30
worker: retry popping before going to sleep

BusyJay:
To avoid frequently sleep and wake up, try again popping several times
before putting thread to sleep.

### DIFF
--- a/src/pool/worker.rs
+++ b/src/pool/worker.rs
@@ -1,7 +1,8 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
 use crate::pool::{Local, Runner};
-use crate::queue::TaskCell;
+use crate::queue::{Pop, TaskCell};
+use parking_lot_core::SpinWait;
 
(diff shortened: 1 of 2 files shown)


src/pool/worker.rs
```diff
@@ -19,21 +20,34 @@ where
...
+        if t.is_some() {
+            self.runner.resume(&mut self.local);
+        }
+        t
+    }
+
     pub fn run(mut self) {
         self.runner.start(&mut self.local);
         while !self.local.core().is_shutdown() {
             let task = match self.local.pop() {
```
sticnarf: Can we just use the implementation of `slow_pop` here and skip this `pop`?

BusyJay: Just to avoid additional function jump.

sticnarf: Even now the compiler seems to inline `slow_pop`. (`slow_pop` does not apppear in symbol list)

sticnarf (approved):
LGTM

AndreMouche (approved):
LGTM

PR #23, opened 2019-12-30
queue: unify naming

BusyJay:
Change simple queue to work stealing queue, and use `TaskInjector`
instead of `QueueInjector`, `LocalQueue` instead of `QueueLocal`.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -11,7 +11,7 @@
 pub mod multilevel;
 
 mod extras;
-mod simple;
+mod work_stealing;
 
 pub use self::extras::Extras;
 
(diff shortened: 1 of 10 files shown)


sticnarf:
I'm a bit worried that when we have a `work_stealing` queue, people may misunderstand the other queue is not work stealing .

BusyJay:
Do you have any recommendation? How about `SingleLevel`?

sticnarf:
`SingleLevel` sounds fine :)

src/queue.rs
```diff
@@ -27,7 +27,7 @@ pub trait TaskCell {
 pub struct TaskInjector<T>(InjectorInner<T>);
 
 enum InjectorInner<T> {
-    WorkStealing(work_stealing::TaskInjector<T>),
+    WorkStealing(single_level::TaskInjector<T>),
```
sticnarf: Enum variant name needs updating.

sticnarf (approved):
LGTM

AndreMouche (approved):
LGTM

PR #24, opened 2019-12-31
task/future: support spawning locally

BusyJay:
It uses a thread local variable to detect if it's in thread pool.
Spawning locally can utilize maximum cache locality.

### DIFF
--- a/src/task/future.rs
+++ b/src/task/future.rs
@@ -5,6 +5,7 @@
 use crate::pool::{Local, Remote};
 use crate::queue::Extras;
 
+use std::borrow::Cow;
 use std::cell::{Cell, UnsafeCell};
 use std::future::Future;
 use std::mem::ManuallyDrop;
@@ -107,13 +108,13 @@ unsafe fn drop_raw(this: *const ()) {
     drop(task_cell(this as *const Task))
 }
 
-unsafe fn wake_impl(task_cell: &TaskCell) {
-    let task = &task_cell.0;
-    let mut status = task.status.load(SeqCst);
+unsafe fn wake_impl(task: Cow<'_, Arc<Task>>) {
+    let mut status = task.as_ref().status.load(SeqCst);
     loop {
         match status {
             IDLE => {
                 match task
+                    .as_ref()
                     .status
                     .compare_exchange_weak(IDLE, NOTIFIED, SeqCst, SeqCst)
                 {
@@ -126,6 +127,7 @@ unsafe fn wake_impl(task_cell: &TaskCell) {
             }
             POLLING => {
                 match task
+                    .as_ref()
                     .status
                     .compare_exchange_weak(POLLING, NOTIFIED, SeqCst, SeqCst)
                 {
@@ -141,13 +143,13 @@ unsafe fn wake_impl(task_cell: &TaskCell) {
 #[inline]
 unsafe fn wake_raw(this: *const ()) {
     let task_cell = task_cell(this as *const Task);
-    wake_impl(&task_cell);
+    wake_impl(Cow::Owned(task_cell.0));
 }
 
 #[inline]
 unsafe fn wake_ref_raw(this: *const ()) {
     let task_cell = ManuallyDrop::new(task_cell(this as *const Task));
-    wake_impl(&task_cell);
+    wake_impl(Cow::Borrowed(&task_cell.0));
 }
 
 #[inline]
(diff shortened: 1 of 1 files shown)


src/task/future.rs
```diff
@@ -147,10 +149,57 @@ unsafe fn task_cell(task: *const Task) -> TaskCell {
 #[inline]
 unsafe fn clone_task(task: *const Task) -> TaskCell {
     let task_cell = task_cell(task);
+    let extras = { &mut *task_cell.0.extras.get() };
+    // `remote` is none only when it has been constructed but never been polled.
```
sticnarf: On the first poll, remote also seems None.

src/task/future.rs
```diff
@@ -151,6 +164,47 @@ unsafe fn clone_task(task: *const Task) -> TaskCell {
...
+    /// Local queue reference that is set before polling and unset after polled.
+    static LOCAL: Cell<*mut Local<TaskCell>> = Cell::new(std::ptr::null_mut());
+}
+
+unsafe fn wake_task(task: Cow<'_, Arc<Task>>, reschedule: bool) {
+    LOCAL.with(|ptr| {
+        if ptr.get().is_null() {
+            // It's out of polling process, has to be spawn to global queue.
+            // It needs to clone to make it safe as it's unclear whether `self`
+            // is still used inside method `spawn` after `TaskCell` is dropped.
```
sticnarf, line 168: Strictly speaking, there is still some risk because `Waker` is accidentally `Sync`. https://github.com/rust-lang/rust/issues/66481
It means the user might be able to move the reference to other threads (for example, using crossbeam::scoped`) and break our assumption.

However, I think it can be ignored. Maybe we can give some documentation and use `expect` to give some messages?

src/task/future.rs
```diff
@@ -85,6 +88,23 @@ unsafe fn waker(task: *const Task) -> Waker {
 #[inline]
 unsafe fn clone_raw(this: *const ()) -> RawWaker {
     let task_cell = clone_task(this as *const Task);
+    let extras = { &mut *task_cell.0.extras.get() };
```
sticnarf: I think we should use an immutable reference here and get the mutable reference only at L98 to make the program sound.

src/task/future.rs
```diff
@@ -151,6 +172,47 @@ unsafe fn clone_task(task: *const Task) -> TaskCell {
...
+    LOCAL.with(|ptr| {
+        if ptr.get().is_null() {
+            // It's out of polling process, has to be spawn to global queue.
+            // It needs to clone to make it safe as it's unclear whether `self`
+            // is still used inside method `spawn` after `TaskCell` is dropped.
+            (*task.as_ref().extras.get())
+                .remote
+                .as_ref()
+                .expect("waker should not be moved to other threads by reference")
+                .spawn(TaskCell(task.clone().into_owned()));
```
sticnarf, line 173: Sorry I still don't understand why we need a clone here...

src/task/future.rs
```diff
@@ -427,29 +483,62 @@ mod tests {
         assert_eq!(res_rx.recv().unwrap(), 4);
     }
 
+    struct ForwardWaker {
```
sticnarf: It's identical to `WakeLater`?

src/task/future.rs
BusyJay: Why is it unsound?

src/task/future.rs
BusyJay: TaskCell can be dropped inside method `spawn`, which can make self invalid.

src/task/future.rs
BusyJay: Exactly. Didn't notice it before.

src/task/future.rs
sticnarf: Two wakers in different threads concurrently, then there can be two multable references to the extras at the same time.

BusyJay: If you mean `clone_raw` can be executed concurrently due to L100, then it's unsound either to move it to L98. Lock is required.

sticnarf: Ahh, you're right. Executing L98 and L91 concurrently is unsound.

src/task/future.rs
```diff
@@ -257,6 +240,15 @@ impl crate::pool::Runner for Runner {
                     task.status.store(COMPLETED, SeqCst);
                     return true;
                 }
+                let extras = { &mut *task.extras.get() };
```
sticnarf, line 243: I think it should be executed before poll. Otherwise, before the first poll finished, waking a waker moved to other threads will panic.

sticnarf: We can use `mut_extras` here.

BusyJay: No, it won't. Waking a polling future will mark it NOTIFIED instead of actually waking it up.

BusyJay: > We can use mut_extras here

They are not the same `extras`. Besides, task_cell is moved.

sticnarf: Ah. I got it.

sticnarf (approved):
LGTM

breezewish (approved):

PR #25, opened 2020-01-02
*: make spawn easier

BusyJay:
This PR adds a new trait `WithExtras` to make spawn easier. Now spawning futures with default extras in future pool can be as simple as
```
pool.spawn(f);
```
rather than in the past
```
pool.spawn(TaskCell::new(f, queue::extras()));
```
.

The old way is not removed as user may want to provide a custom extras.

This PR also make the queue types private. Those types can only be used when constructing pool, it's meaningless and distracting to expose the full type.

### DIFF
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -3,7 +3,7 @@
 use crate::pool::spawn::QueueCore;
 use crate::pool::worker::WorkerThread;
 use crate::pool::{CloneRunnerBuilder, Local, Remote, Runner, RunnerBuilder, ThreadPool};
-use crate::queue::{self, multilevel, LocalQueue, TaskCell};
+use crate::queue::{self, LocalQueue, QueueType, TaskCell};
 use crate::task::{callback, future};
 use std::sync::{Arc, Mutex};
 use std::thread;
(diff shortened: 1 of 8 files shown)


src/pool/builder.rs
```diff
@@ -178,36 +178,21 @@ impl Builder {
...
-    /// number of local queues and returns the task injector and local queues.
-    ///
-    /// In some cases, especially building up a large application, a task
-    /// scheduler is required before spawning new threads. You can use this
-    /// to separate the construction and starting.
-    pub fn freeze_with_queue<T>(
-        &self,
-        queue_builder: impl FnOnce(usize) -> (TaskInjector<T>, Vec<LocalQueue<T>>),
-    ) -> (Remote<T>, LazyBuilder<T>)
+    pub fn freeze<T>(&self, multilevel: Option<multilevel::Builder>) -> (Remote<T>, LazyBuilder<T>)
```
sticnarf: I don't think it's a good idea to put a `multilevel` parameter here. It is not extensible at all. 

If we really don't want to expose `TaskInjector` and `LocalQueue`, can we give this builder a trait while still not exposing the queue types? (Rustdoc will show them as black and the user has no ability to implement the trait)

BusyJay: How about using a unify builder?

sticnarf: Then, in order not to expose the queue type, we will have a builder with no `build` function visible to users (this pr is the same)...It seems a bit strange to me but acceptable.

sticnarf (approved):
LGTM

breezewish (approved):

PR #26, opened 2020-01-03
*: add benchmark

BusyJay:
The PR introduces four benchmark cases and compare them with tokio,
threadpool and async-std.

Benchmark shows that there is still improvment room for yatp in serveral
cases, especially for future pool.

Only single level queue is benched as workload works for multilevel is
quite different from single level. It can be added in the future.

### DIFF
--- a/benches/chained_spawn.rs
+++ b/benches/chained_spawn.rs
@@ -1,0 +1,156 @@
+// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
+
+use criterion::*;
+
+mod yatp_callback {
+    use criterion::*;
+    use std::sync::mpsc;
+    use yatp::task::callback::Handle;
+
+    pub fn chained_spawn(b: &mut Bencher<'_>, iter_count: usize) {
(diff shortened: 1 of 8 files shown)


benches/chained_spawn.rs
```diff
@@ -0,0 +1,156 @@
...
+                    iter(s2, done_tx, n - 1);
+                });
+            }
+        }
+
+        let (done_tx, done_rx) = mpsc::sync_channel(1000);
+
+        b.iter(move || {
+            let done_tx = done_tx.clone();
+            let remote = pool.remote().clone();
```
sticnarf, line 60: Now `ThreadPool::remote()` returns `Remote`. I think we can change it to `&Remote`.

sticnarf (approved):
LGTM.
An obvious optimization is pushing the task into the local queue if the spawn happens in the working thread. We did it wrong in putting the optimization in the waker. Instead, we should put the logic into `Remote::spawn`.

AndreMouche (approved):
LGTM

PR #27, opened 2020-01-06
add basic metrics of multilevel task queue

sticnarf:
It's quite useful to get how long tasks use and the chance of being scheduled of each level to investigate issues about the effect of multilevel feedback scheduling.

### DIFF
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -434,8 +434,8 @@ pub struct Config {
 impl Config {
     /// Sets the name of the multilevel task queue. Metrics of multilevel
     /// task queues are available if name is provided.
-    pub fn name(mut self, name: Option<String>) -> Self {
-        self.name = name;
+    pub fn name(mut self, name: Option<impl Into<String>>) -> Self {
+        self.name = name.map(Into::into);
         self
     }
 
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -41,9 +41,16 @@ const ADJUST_CHANCE_INTERVAL_US: i64 = 1_000_000;
 /// exceeds this value, level chances need to be adjusted.
 const ADJUST_CHANCE_THRESHOLD: f64 = 0.05;
 
+/// The initial chance that a level 0 task is scheduled.
+///
+/// The value is not so important because the actual chance will be adjusted
+/// according to the real-time workload.
 const INIT_LEVEL0_CHANCE: f64 = 0.8;
+
 const MIN_LEVEL0_CHANCE: f64 = 0.5;
 const MAX_LEVEL0_CHANCE: f64 = 0.98;
+
+/// The amount that the level 0 chance is increased or decreased each time.
 const ADJUST_AMOUNT: f64 = 0.06;
 
 /// The injector of a multilevel task queue.
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -65,6 +65,14 @@ where
         let level = task_cell.mut_extras().current_level as usize;
         self.level_injectors[level].push(task_cell);
     }
+
+    pub(super) fn statistics(&self) -> MultilevelStatistics {
+        MultilevelStatistics {
+            level0_elapsed: self.manager.level0_elapsed.as_duration(),
+            total_elapsed: self.manager.total_elapsed.as_duration(),
+            level0_chance: self.manager.get_level0_chance() as f64 / u32::max_value() as f64,
+        }
+    }
 }
 
 /// The local queue of a multilevel task queue.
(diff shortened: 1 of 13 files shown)


BusyJay:
Why not use rust-prometheus instead? In that case, users of the library does even have to make their own wrappers.

sticnarf:
@BusyJay We can't assume the user uses prometheus.

BusyJay:
They can still access the data via prometheus's APIs.

/cc @breeswish, is it easy for other metrics implement to collaborate with rust-prometheus?

sticnarf:
I think it mightn't be a good idea to use prometheus-rust in a library. Because the registry needs to interact with the metrics. If the version of the registry differs from that of the metrics, I believe it's likely to be true, then using prometheus in the library cannot bring any benefit.

BusyJay:
Why need to use registry? What I mean is use prometheus' counters and histograms. It should behave like rocksdb's metrics: https://github.com/facebook/rocksdb/blob/master/include/rocksdb/statistics.h.

sticnarf:
Then what's the benefit of using prometheus's counters, just an atomic wrapper?
If the versions don't match, the counters and histograms cannot be used with the prometheus registry that the user uses and it doesn't bring any more convenience. Why bother introduce a library only for its counter type?

sticnarf:
It is easy for us to unify the version of rust-prometheus used by tikv and this library. But it doesn't change that it's generally not a good idea to use rust-prometheus in a library.

BusyJay:
> Then what's the benefit of using prometheus's counters, just an atomic wrapper?

The benefit is that:
1. we don't need to define our own types, such as counter, histogram or registry. For example, to check the performance of a benchmark, we can dump the metrics easily, instead of building a struct and customize its output from scratch;
2. very easy to integrate with other system, all need to be done are just `registry.register`.

> If the versions don't match, the counters and histograms cannot be used with the prometheus registry 

I think it's very similar with log crate. log crate also has similar layout but it doesn't prevent it from being widely used by both libraries and binaries. I don't think versions are the defeater of using another library.

sticnarf:
Ok. I can change to use prometheus if you have strong opinion.

PS: The log crate uses a trick to deal with compatibility. log 0.3 depends on log 0.4 so the types can be used across versions. We can apply this trick on rust-prometheus too, I think. /cc @breeswish

sticnarf:
@BusyJay PTAL again.
I choose to use global static MetricsVec so if the user uses several thread pools at the same time, it is convenient to register the vec only once and get information of all thread pools.

src/queue/multilevel.rs
```diff
@@ -266,57 +286,67 @@ impl LevelManager {
...
     fn maybe_adjust_chance(&self) {
         if self.adjusting.compare_and_swap(false, true, SeqCst) {
             return;
         }
         // The statistics may be not so accurate because we cannot load two
         // atomics at the same time.
-        let total = self.total_elapsed.0.load(SeqCst);
-        if Duration::from_micros(total) < ADJUST_CHANCE_INTERVAL {
-            // Another thread just adjusted the chances.
+        let total = self.total_elapsed_us.get();
```
BusyJay, line 302: I think the order will be `Relaxed` for metrics, is it OK?

sticnarf: I think it's OK as soon as the time of each handling is not very long so that the inaccuracy doesn't have much effect.

BusyJay: I'm worry that `total_diff` can be zero as reorder is allowed now.

sticnarf: If `total_diff` is zero, the function returns in L297. The calculation is different here and there is no risk of div by 0 like before. Anything else worrying?

src/queue/multilevel.rs
```diff
@@ -266,57 +286,67 @@ impl LevelManager {
...
+        let total_diff = total - self.last_total_elapsed_us.get();
+        if total_diff < ADJUST_CHANCE_INTERVAL_US {
```
BusyJay, line 304: So the metrics report interval should be less than `ADJUST_CHANCE_INTERVAL_US`, otherwise it can be always true.

sticnarf: Sorry I don't get what you mean. Reporting the metrics shoudn't reset its value and `last_total_elapsed_us` is only set at L304.

BusyJay: Oh, turns out I misunderstood the processing of metrics.

BusyJay (dismissed):

BusyJay:
@AndreMouche @breeswish PTAL

src/queue/multilevel.rs
```diff
@@ -26,20 +28,23 @@ const CHANCE_RATIO: u32 = 4;
...
 /// exceeds this value, level chances need to be adjusted.
 const ADJUST_CHANCE_THRESHOLD: f64 = 0.05;
 
-// The chance values below are the numerators of fractions with u32::max_value()
-// as the denominator.
-const INIT_LEVEL0_CHANCE: u32 = 3_435_973_836; // 0.8
-const MIN_LEVEL0_CHANCE: u32 = 1 << 31; // 0.5
-const MAX_LEVEL0_CHANCE: u32 = 4_209_067_949; // 0.98
-const ADJUST_AMOUNT: u32 = (MAX_LEVEL0_CHANCE - MIN_LEVEL0_CHANCE) / 8; // 0.06
+const INIT_LEVEL0_CHANCE: f64 = 0.8;
```
AndreMouche, line 48: Could we add some description about this value?

AndreMouche (dismissed):
LGTM

src/queue/multilevel.rs
sticnarf: I add some comments to it. PTAL

BusyJay (approved):

AndreMouche (approved):
LGTM

PR #28, opened 2020-01-06
*: introduce coarse time

BusyJay:
Scheduling algorithm doesn't require accurate time measurement, so use
coarse time to reduce overhead. Benchmark shows that there are about 10%
improvement in all cases.


sticnarf:
For the same reason stated here (https://github.com/tikv/tikv/pull/6375#discussion_r363340565), I worry using coarse time for `schedule_time` will lead to unexpected high latency.

I don't think it is necessary to chance the cleanup time to coarse time. `recent()` is right for reducing the cost of `Instant::now()`.

PR #29, opened 2020-01-08
spawn through local queue if possible

sticnarf:
In this PR, if `spawn` happens in tne working threads, tasks are pushed into the local queue for locality. It also removes the local wake mechanism of futures. `Remote` is renamed to `Handle` for its new meaning.

Benchmark shows up to 100% improvement on chain_spawn.
But there is a regression on ping_pong  when `ping_count` is large. This is because it is much more expensive to increase the capacity of crossbeam `Worker` (allocating new buffer and copying all) than `Injector` (just adding a new block).

### DIFF
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -2,7 +2,7 @@
 
 use crate::pool::spawn::QueueCore;
 use crate::pool::worker::WorkerThread;
-use crate::pool::{CloneRunnerBuilder, Local, Remote, Runner, RunnerBuilder, ThreadPool};
+use crate::pool::{CloneRunnerBuilder, Handle, Local, Runner, RunnerBuilder, ThreadPool};
 use crate::queue::{self, LocalQueueBuilder, QueueType, TaskCell};
 use crate::task::{callback, future};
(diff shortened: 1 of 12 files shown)


src/pool/builder.rs
```diff
@@ -46,13 +46,13 @@ impl Default for SchedConfig {
 }
 
 /// A builder for lazy spawning.
-pub struct LazyBuilder<T> {
+pub struct LazyBuilder<T: Send + 'static> {
```
BusyJay: Better add the constrain only when it's necessary.

src/pool/spawn.rs
```diff
@@ -133,21 +137,60 @@ impl<T: TaskCell + Send> QueueCore<T> {
     }
 }
 
+thread_local! {
+    static LOCAL_INJECTOR: Cell<TlsLocalInjector> = Cell::new(TlsLocalInjector::uninit());
+}
+
+static POOL_ID_ALLOCATOR: AtomicU64 = AtomicU64::new(1);
```
BusyJay: You can just use the address of `Arc<QueueCore>`.

PR #35, opened 2020-04-15
calculate appropriate level 0 chance

sticnarf:
## Problem

Currently the maximum level 0 chance is 0.98. Sometimes we still fail to achieve our target even with this maximum value. For example, if we have plenty of tasks of all levels to run and the average time spent in handling a level 0 task once is 50us while the average time spent in handling other tasks once is 1500us. We can calculate the final time proportion of level 0 is (50 * 0.98) / (50 * 0.98 + 1500 * 0.02) = 0.62, which is far lower than we expected.

## Solution
Firstly, this PR raises the maximum level 0 chance to 0.9999, which should suffice in most cases. And the way we adjust the chance is improved. The new chance is now calculated based on the assumption that we have plenty of tasks in all levels. (It is OK when the assumption is false because then we actually don't need to balance tasks of different levels.

## Formula

Say the average time spent in handling level 1 and 2 tasks once is _k_ times of that for level 0 task, the current time proportion of level 0 tasks is _m_, the target proportion is _m0_ and the chance of level 0 tasks is _p_. We want to get an appropriate new chance of level 0 tasks, _p'_.

So, we have simultaneous equations below:

p/(p+k(1-p)) = m
p'/(p'+k(1-p')) = m0

Solving the equations we can get p' = m0(1-m)p/(m0(1-m0)-(m-m0)(p-(1-m0))).

In our implementation, we ensure 0 < p < 1, 0 <= m <= 1, 0 < m0 < 1. The denominator can be proved to be greater than zero so we needn't check it in our code.

### DIFF
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -94,7 +94,7 @@ where
             );
         }
         ThreadPool {
-            remote: Remote::new(self.core.clone()),
+            remote: Remote::new(self.core),
             threads: Mutex::new(threads),
         }
     }
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -38,10 +38,6 @@ const FLUSH_LOCAL_THRESHOLD_US: i64 = 100_000;
 /// adjust level chances and reset the total elapsed time.
 const ADJUST_CHANCE_INTERVAL_US: i64 = 1_000_000;
 
-/// When the deviation between the target and the actual level 0 proportion
-/// exceeds this value, level chances need to be adjusted.
-const ADJUST_CHANCE_THRESHOLD: f64 = 0.05;
-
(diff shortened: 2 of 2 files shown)


BusyJay:
> the denominator can be proved to be greater than zero

How?

sticnarf:
> > the denominator can be proved to be greater than zero
> 
> How?

if 
image

then 
image

Because m and m_0p are all > 0, p-(1-m0) > 0.
So we have 
image

And also, we know m <= 1, then, 

image

Multiply the denominator in both side, we get
image

But both (1-m0) and (1-p) > 0, so we get a contradiction.
QED

BusyJay (dismissed):
Cool.

BusyJay (approved):

PR #37, opened 2020-05-14
multilevel: use atomic for elapsed time

BusyJay:
This PR brings two updates:

- use atomic variable instead of `IntCounter` to record request time.
- if a request has fixed level, skip recording request time.

Atomic variable is faster to be allocated, which can save some time when request is very small. A simple spawn_many bench shows about 50% improvement.

If a request has fixed level, recorded request time will not have impact on itself, so it can be skipped completely. Although there may be requests share the same task ID, but mixing dynamic and fixed level usage seems odd to me. Bench shows that if `spawn_many` is benched with fixed level, multilevel is almost as fast as single_level.

### DIFF
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -17,7 +17,7 @@ use prometheus::local::LocalIntCounter;
 use prometheus::{Gauge, IntCounter};
 use rand::prelude::*;
 use std::cell::Cell;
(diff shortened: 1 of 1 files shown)


BusyJay:
Benchmark in TiKV shows point get QPS increases by about 4%. Although there is still 1% difference from disabling it completely.

sticnarf (approved):
LGTM

PR #38, opened 2020-05-15
*: adjust benchmark

BusyJay:
100 is hard to be distinguished from 1000 when filtering benchmark case,
so this PR changes the dataset to be 2 based.

A benchmark to multilevel queue is also added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,3 +45,4 @@ harness = false
 
 [profile.bench]
 codegen-units = 1
+debug = true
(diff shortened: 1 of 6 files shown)


sticnarf (commented):
Why is the multilevel benchmark only added to `spawn_many`?

And actually criterion supports using regex to filter tests since 0.3.1. It's possible to run 100 using `cargo bench -- 100$`.

BusyJay:
Thanks for the suggestion. Use simple match is more friendly.

sticnarf (approved):
LGTM

PR #40, opened 2020-08-11
Add clone trait for builder

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>


sticnarf:
What's the motivation of making it `Clone`?

PR #45, opened 2020-10-14
ban local spawns across pools

sticnarf:
Previously we utilize local spawn as long as the TLS handle `LOCAL` exists.

However, when a waker is called in a thread of a different yatp pool, the future can be transfered from one pool to another unexpectedly.

This PR checks whether the waker and the current local handle are of the same pool. If no, we spawn remotely.

### DIFF
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -94,7 +94,7 @@ where
             );
         }
         ThreadPool {
-            remote: Remote::new(self.core.clone()),
+            remote: Remote::new(self.core),
             threads: Mutex::new(threads),
         }
     }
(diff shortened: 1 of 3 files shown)


sticnarf:
PTAL @BusyJay 
cc @Little-Wallace

BusyJay (approved):

zhangjinpeng87:
Any issue can cause if not fixed?

sticnarf:
> Any issue can cause if not fixed?

Not any known issue in TiKV 4.0.

There would be problem if a future in pool A reads TLS specific to pool A and the future is notified in another yatp pool B.

PR #47, opened 2020-10-14
Make futures only hold weak ptr of the core

### LINKED ISSUE #46: Future does not drop after yatp pool has stopped.
### Problem 

* When I run a future which capture some object in yatp pool, I found that these object may live longer than yatp pool. Even if yatp pool stopped, these objects still do not drop. 
* Yatp pool and `Remote` will share one queue by `Arc`. The future in queue only drop when all `Remote` and yatp pool have dropped. But if one `Remote` was captured into a future which running in yatp pool, this operation will cause 'Circular reference'. And all future in queue will never drop.

### Proposal

* Once yatp pool stop,  drop all future in queue immediately.

sticnarf:
Closes #46 

This PR makes `Remote` not hold a strong reference to the core and always check if the core is still alive before any operation.

This adds extra cost that upgrades a weak ptr to a strong Arc.

```
chained_spawn/yatp::future/1024                                                                            
                        time:   [906.10 us 913.73 us 919.75 us]
                        change: [-2.0794% -0.7563% +0.6007%] (p = 0.29 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe
ping_pong/yatp::future/1024                                                                             
                        time:   [1.0532 ms 1.0718 ms 1.0913 ms]
                        change: [-4.8954% -2.4524% +0.2339%] (p = 0.07 > 0.05)
                        No change in performance detected.

     Running target/release/deps/spawn_many-c9d51dfaec34d6eb
spawn_many/yatp::future/1024                                                                             
                        time:   [3.6443 ms 3.7237 ms 3.7983 ms]
                        change: [-2.0927% +0.9957% +4.0339%] (p = 0.52 > 0.05)
                        No change in performance detected.

     Running target/release/deps/yield_many-939c8e996b0eade2
yield_many/yatp::future/1024                                                                             
                        time:   [3.0346 ms 3.0890 ms 3.1437 ms]
                        change: [-1.2660% +1.2049% +3.7731%] (p = 0.34 > 0.05)
                        No change in performance detected.
```

### DIFF
--- a/src/task/future.rs
+++ b/src/task/future.rs
@@ -5,7 +5,6 @@
 use crate::pool::{Local, Remote};
 use crate::queue::{Extras, WithExtras};
 
-use std::borrow::Cow;
 use std::cell::{Cell, UnsafeCell};
 use std::future::Future;
 use std::mem::ManuallyDrop;
@@ -13,6 +12,7 @@ use std::pin::Pin;
 use std::sync::atomic::{AtomicU8, Ordering::SeqCst};
 use std::sync::Arc;
 use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
+use std::{borrow::Cow, ptr, sync::Weak};
 use std::{fmt, mem};
 
 /// The default repoll limit for a future runner. See `Runner::new` for
@@ -184,8 +184,11 @@ unsafe fn wake_task(task: Cow<'_, Arc<Task>>, reschedule: bool) {
             .remote
             .as_ref()
             .expect("remote should exist!!!");
-        let out_of_polling =
-            ptr.get().is_null() || !Arc::ptr_eq(&(*ptr.get()).core(), &task_remote.core);
+        let out_of_polling = ptr.get().is_null()
+            || !ptr::eq(
+                Arc::as_ptr(&(*ptr.get()).core()),
+                Weak::as_ptr(&task_remote.core),
+            );
         if out_of_polling {
             // It's out of polling process, has to be spawn to global queue.
             // It needs to clone to make it safe as it's unclear whether `self`
--- a/src/task/future.rs
+++ b/src/task/future.rs
@@ -2,7 +2,7 @@
 
 //! A [`Future`].
 
-use crate::pool::{Local, Remote};
+use crate::pool::{Local, WeakRemote};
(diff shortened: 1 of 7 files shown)


BusyJay:
Why not just drain the queue when the last worker thread exits?

sticnarf:
> Why not just drain the queue when the last worker thread exits?

@BusyJay  The hardest problem is not draining the existent tasks in the queue. It's to avoid wakers to push new tasks to the injector.

And a simple atomic read cannot avoid it completely. Because writing such a flag can happen between the check and pushing the task to the queue. That's why we need `Weak`.

BusyJay:
How about creating a WeakRemote for TaskCell that only contains weak pointer to remote? In such way, there will be no circular references. Although it still requires all Remotes and pools be dropped before releasing all the futures.

/cc @Little-Wallace can this solve your problem?

Little-Wallace:
@BusyJay  But one `Remote` may be capture to a future.....

BusyJay:
Yes. In my opinion, a thread pool hardly be shutdown and recreated during runtime. The library is designed under an assumption that threads are allowed to live long. We can try our best effort to avoid leak, but should optimize for the cases we focus.

src/task/future.rs
```diff
@@ -163,7 +163,7 @@ unsafe fn clone_task(task: *const Task) -> TaskCell {
     let extras = { &mut *task_cell.0.extras.get() };
     if extras.remote.is_none() {
         LOCAL.with(|l| {
-            extras.remote = Some((&*l.get()).remote());
+            extras.remote = Some((&*l.get()).remote().downgrade());
```
BusyJay: Why not use `weak_remote` directly?

BusyJay (commented):
Rest LGTM

BusyJay:
How does it perform when using `WeakRemote`?

sticnarf:
> How does it perform when using `WeakRemote`?

I update the benchmark result in the PR description. We don't have remote spawns via future notifies in the benchmark, so there is not much difference.

BusyJay (approved):

PR #49, opened 2021-04-12
using compare_exchange instead of deprecated compare_and_swap

yaozongyou:
Signed-off-by: Yao Zongyou <yaozongyou@vip.qq.com>


BusyJay (approved):
Thanks!

PR #52, opened 2021-04-19
Use GitHub Actions to test and bench

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

You can check some examples from here: https://github.com/JmPotato/yatp/actions


JmPotato:
cc @BusyJay @sticnarf

JmPotato:
It looks like we need to remove Travis from the CI settings of this repo and move to GitHub Actions. BTW, GitHub Actions is way faster than Travis now.

sticnarf (approved):
Looks good. We may need @BusyJay to configure the repo.

.github/workflows/bench.yaml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust_version }}
+          override: true
+      - name: Bench
+        run: cargo bench --all -- --test
```
BusyJay, line 36: Isn't it covered by test.yaml already?

sticnarf: No. `cargo test` does not run benchmarks as tests.

BusyJay (approved):

PR #57, opened 2021-09-10
Run address sanitizer on CI

sticnarf:
A lot of unsafe code is used to implement a future executor. It is better to add the address sanitizer to CI to reduce possible memory errors.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -97,4 +97,4 @@ jobs:
           toolchain: nightly
           override: true
       - name: Run address sanitizer on tests
-        run: cargo test --target x86_64-unknown-linux-gnu --all --all-features -- --nocapture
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -81,3 +81,20 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
+      - name: Run address sanitizer on tests
+        run: cargo test --target x86_64-unknown-linux-gnu --all --all-features -- --nocapture
```
BusyJay: I can see the documentation says std should also be rebuilt. https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html

sticnarf: Now I added `-Zbuild-std`.

BusyJay (approved):

Issue #60, opened 2021-12-01
Support dynamically scaling workers of the pool


ethercflow:
### Describe the feature you'd like:
I want to adjust the number of workers online, Actual scenarios such as scaling `read pool` (controlled by `unified.max-thread-count`) of [TiKV](https://github.com/tikv/tikv) without restart. Expect to reduce performance jitter as much as possible during the adjustment process.

### Technical proposal
- Add a new member `core_thread_count` in `SchedConfig`,  the naming of core borrows `corePoolSize` from [ThreadPoolTaskExecutor](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/concurrent/ThreadPoolTaskExecutor.html). 
In our scenario, `max_thread_count` represents the maximum scale value of the thread pool, 
`core_thread_count` represents the number of actual participation in `Remote<T>` scheduling. If `core_thread_count` is not set during `build`, it will be equal to the actual set `max_thread_count`.
- Add a new method `runnable` in `Local<T>`, it will call `need_to_park_worker` which is added in `QueueCore<T>` to To determine whether it can run, if not, call `sleep` which is added in `Local<T>` to park self with addr `QueueCore<T> | 1`, so it will not participate in `Remote<T>` scheduling
- Add `scale_workers` in `Remote<T>`, When scaling up, the number of workers needed to unpark from the queue with addr `QueueCore<T> | 1`. When scaling down, unpark addr `QueueCore<T>`, so that threads in `pop_and_sleep` can also participate in recycling （by call `runnable`  mentioned above）
- Add `scale_workers`  in `ThreadPool<T>` too, it just calls `Remote<T>`'s `scale_workers`.
- Add a new member `core_workers` in `QueueCore<T>`, as the value of `active_workers` changes with the working status of workers, it cannot be reused. It is also an atomic variable and is kept in sync by using `compare_exchange_weak`. This member is used in `scale_workers` and `Runnable`.

#### Pros

1. The adjustment of the number of threads is very lightweight, and there are no thread creation and destruction operations and lock operations
2. Without scale, judging whether it is runnable is just comparing two atomic variables

#### Cons

1. During `build`, `max_thread_count` thread objects will be created at once, there will be a little more CPU overhead and memory overhead

### Process

- The feature has been implemented, under the stability test.

PR #61, opened 2021-12-01
Support dynamically scaling workers of the pool (#60)

ethercflow:
This Feature Request description and Technical proposal can be seen here: #60 

Tests that have been run, the test env is 
- IDC physical machine (CentOS Linux release 8.4.2105/4.18.0-305.10.2.el8_4.x86_64)
- WSL2 (AMD bullseye/sid/5.15.0-microsoft-standard-WSL2)

Testcases:

``` sh
#!/bin/bash -e

while true
do
        cargo bench --all -- --test
        cargo bench --all --all-features -- --test
        cargo test --all -- --nocapture
        cargo test --all --all-features -- --nocapture
done
```

- Ran `cargo bench --all -- --test` in a shell loop, will exist if  failed: passed with 26 success, 4 ignored, tested 50+ times

``` sh
test queue::multilevel::tests::test_get_elapsed_deadlock ... ignored
test task::future::tests::test_no_preemptive_task ... ignored
test task::future::tests::test_repoll_limit ... ignored
test task::future::tests::test_reschedule ... ignored
```

- Ran `cargo bench --all --all-features -- --test` in a shell loop, will exist if  failed: passed with 30(all) success. tested 50+ times

- I wrote a system testcase: [test-yatp](https://github.com/ethercflow/test-yatp), and had ran for a few hours. 

The test logic is simple:

``` rust
fn scale_workers() {
    let pool = Builder::new("SP")
        .max_thread_count(16)
        .core_thread_count(4)
        .build_callback_pool();
    let handler = pool.remote().clone();
    let builder = thread::Builder::new().name("wl".to_string());
    builder
        .spawn(move || {
            loop {
                let (tx, rx) = mpsc::channel();
                // A bunch of tasks should be executed correctly.
                let cases: Vec<_> = (10..100000000).collect();
                for id in &cases {
                    let t = tx.clone();
                    let id = *id;
                    handler.spawn(move |_: &mut Handle<'_>| t.send(id).unwrap());
                }
                let mut ans = vec![];
                for _ in 10..100000000 {
                    let r = rx.recv_timeout(Duration::from_secs(1)).unwrap();
                    ans.push(r);
                }
                ans.sort();
                assert_eq!(cases, ans);
                println!("finish one loop");
            }
        })
        .unwrap();
    loop {
        let mut rng = rand::thread_rng();
        let new_thread_count = rng.gen_range(1..16);
        println!("scale workers to {}", new_thread_count);
        pool.remote().scale_workers(new_thread_count);
        thread::sleep(Duration::from_secs(10));
    }
}
```

While running, execute the `pidstat -G yatp -t 1` command to observe that the number of threads running is in line with expectations:

``` sh
➜  test-yatp git:(main) ./target/release/test-yatp
scale workers to 8
scale workers to 14
scale workers to 9
```

``` sh
00:57:19      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:20        0   3950591         -  785.00   77.00    0.00    0.00  862.00    17  test-yatp
00:57:20        0         -   3950592   89.00    6.00    0.00    0.00   95.00     6  |__SP-0
00:57:20        0         -   3950593   87.00    8.00    0.00    0.00   95.00     2  |__SP-1
00:57:20        0         -   3950594   90.00    6.00    0.00    0.00   96.00     7  |__SP-2
00:57:20        0         -   3950595   87.00    9.00    0.00    0.00   96.00    10  |__SP-3
00:57:20        0         -   3950604   88.00    8.00    0.00    0.00   96.00    39  |__SP-12
00:57:20        0         -   3950605   88.00    8.00    0.00    0.00   96.00    31  |__SP-13
00:57:20        0         -   3950606   87.00    8.00    0.00    0.00   95.00    33  |__SP-14
00:57:20        0         -   3950607   87.00    8.00    0.00    0.00   95.00    21  |__SP-15
00:57:20        0         -   3950608   83.00   16.00    0.00    0.00   99.00    25  |__wl

00:57:20      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:21        0   3950591         -  797.00  149.00    0.00    0.00  946.00    17  test-yatp
00:57:21        0         -   3950592   75.00   13.00    0.00    0.00   88.00    21  |__SP-0
00:57:21        0         -   3950593   76.00   10.00    0.00    0.00   86.00     7  |__SP-1
00:57:21        0         -   3950594   75.00   11.00    0.00    1.00   86.00    31  |__SP-2
00:57:21        0         -   3950595   75.00   11.00    0.00    0.00   86.00    10  |__SP-3
00:57:21        0         -   3950596   21.00    4.00    0.00    0.00   25.00    29  |__SP-4
00:57:21        0         -   3950597   19.00    5.00    0.00    0.00   24.00    20  |__SP-5
00:57:21        0         -   3950598   20.00    5.00    0.00    0.00   25.00    33  |__SP-6
00:57:21        0         -   3950599   19.00    5.00    0.00    0.00   24.00    13  |__SP-7
00:57:21        0         -   3950600   20.00    4.00    0.00    0.00   24.00    35  |__SP-8
00:57:21        0         -   3950602   20.00    4.00    0.00    0.00   24.00     9  |__SP-10
00:57:21        0         -   3950604   77.00   10.00    0.00    0.00   87.00    39  |__SP-12
00:57:21        0         -   3950605   75.00   11.00    0.00    0.00   86.00     3  |__SP-13
00:57:21        0         -   3950606   75.00   12.00    0.00    0.00   87.00     1  |__SP-14
00:57:21        0         -   3950607   75.00   11.00    0.00    0.00   86.00    23  |__SP-15
00:57:21        0         -   3950608   69.00   32.00    0.00    0.00  101.00    25  |__wl

00:57:34      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:35        0   3950591         -  829.00  100.00    0.00    0.00  929.00    21  test-yatp
00:57:35        0         -   3950592   84.00    8.00    0.00    0.00   92.00    30  |__SP-0
00:57:35        0         -   3950594   83.00   10.00    0.00    0.00   93.00    15  |__SP-2
00:57:35        0         -   3950595   85.00    8.00    0.00    0.00   93.00    17  |__SP-3
00:57:35        0         -   3950597   83.00    9.00    0.00    1.00   92.00     0  |__SP-5
00:57:35        0         -   3950599   85.00    7.00    0.00    0.00   92.00    29  |__SP-7
00:57:35        0         -   3950600   82.00   10.00    0.00    0.00   92.00    34  |__SP-8
00:57:35        0         -   3950602   83.00    9.00    0.00    0.00   92.00    25  |__SP-10
00:57:35        0         -   3950606   85.00    8.00    0.00    0.00   93.00    23  |__SP-14
00:57:35        0         -   3950607   82.00   10.00    0.00    1.00   92.00     1  |__SP-15
00:57:35        0         -   3950608   78.00   19.00    0.00    0.00   97.00    31  |__wl
^C
```

 It does make sense?

Signed-off-by: Wenbo Zhang <ethercflow@gmail.com>

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -34,6 +34,15 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
+    /// Scale workers of the thread pool, the new thread count should be
+    /// smaller than the max_thread_count, otherwise it will become no-op
+    /// and return false.
+    ///
+    /// If the pool is shutdown, it becomes no-op.
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
+        self.remote.scale_workers(new_thread_count)
+    }
+
     /// Shutdowns the pool.
     ///
     /// Closes the queue and wait for all threads to exit.
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -5,15 +5,22 @@ use crate::pool::worker::WorkerThread;
 use crate::pool::{CloneRunnerBuilder, Local, Remote, Runner, RunnerBuilder, ThreadPool};
 use crate::queue::{self, multilevel, LocalQueue, QueueType, TaskCell};
 use crate::task::{callback, future};
-use std::sync::{Arc, Mutex};
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc, Mutex,
+};
 use std::thread;
 use std::time::Duration;
 
 /// Configuration for schedule algorithm.
-#[derive(Clone)]
 pub struct SchedConfig {
     /// The maximum number of running threads at the same time.
     pub max_thread_count: usize,
+    /// The core number of running threads at the same time. It
+    /// is defined as the AtomicUsize because it needs to be used
+    /// to scale the number of workers at runtime, and the adjustable
+    /// range is min_thread_count to max_thread_count.
+    pub core_thread_count: AtomicUsize,
     /// The minimum number of running threads at the same time.
     pub min_thread_count: usize,
     /// The maximum tries to rerun an unfinished task before pushing
@@ -35,12 +42,28 @@ impl Default for SchedConfig {
     fn default() -> SchedConfig {
         SchedConfig {
             max_thread_count: num_cpus::get(),
+            core_thread_count: AtomicUsize::new(0),
             min_thread_count: 1,
             max_inplace_spin: 4,
             max_idle_time: Duration::from_millis(1),
             max_wait_time: Duration::from_millis(1),
             wake_backoff: Duration::from_millis(1),
             alloc_slot_backoff: Duration::from_millis(2),
+        }
+    }
+}
+
+impl Clone for SchedConfig {
+    fn clone(&self) -> Self {
+        SchedConfig {
+            max_thread_count: self.max_thread_count,
+            min_thread_count: self.min_thread_count,
+            core_thread_count: AtomicUsize::new(self.core_thread_count.load(Ordering::SeqCst)),
+            max_inplace_spin: self.max_inplace_spin,
+            max_idle_time: self.max_idle_time,
+            max_wait_time: self.max_wait_time,
+            wake_backoff: self.wake_backoff,
+            alloc_slot_backoff: self.alloc_slot_backoff,
         }
     }
 }
@@ -134,6 +157,19 @@ impl Builder {
         self
     }
 
+    /// Sets the number of threads that can participate in being scheduled
+    /// by `Remote` at the same time. It will be checked when building the
+    /// queue, if this value exceeds `max_thread_count`, it will be adjusted
+    /// to `max_thread_count`.
+    pub fn core_thread_count(&mut self, count: usize) -> &mut Self {
+        if count > 0 {
+            self.sched_config
+                .core_thread_count
+                .store(count, Ordering::SeqCst);
+        }
+        self
+    }
+
     /// Sets the maximum tries to rerun an unfinished task before pushing
     /// back to queue.
     pub fn max_inplace_spin(&mut self, count: usize) -> &mut Self {
@@ -204,6 +240,12 @@ impl Builder {
         T: TaskCell + Send,
     {
         assert!(self.sched_config.min_thread_count <= self.sched_config.max_thread_count);
+        let core_thread_count = self.sched_config.core_thread_count.load(Ordering::SeqCst);
+        if core_thread_count == 0 || core_thread_count > self.sched_config.max_thread_count {
+            self.sched_config
+                .core_thread_count
+                .store(self.sched_config.max_thread_count, Ordering::SeqCst);
+        }
         let (injector, local_queues) = queue::build(queue_type, self.sched_config.max_thread_count);
         let core = Arc::new(QueueCore::new(injector, self.sched_config.clone()));
 
--- a/src/pool/runner.rs
+++ b/src/pool/runner.rs
@@ -12,9 +12,15 @@ use crate::pool::Local;
 /// ```text
 ///   start
 ///     |
-///     | <--- resume
-///     |        |
-///   handle -> pause
+///     | <---------------
+///     |                |
+///   reentrant_start    |
+///     |                |
+///     | <--- resume    | if selected durning scaling up
+///     |        |       |
+///   handle -> pause    |
+///     |                |
+///   reentrant_end ------
 ///     |
 ///    end
 /// ```
@@ -30,6 +36,13 @@ pub trait Runner {
     /// It's guaranteed to be the first method to call before anything else.
     fn start(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
+    /// Called when the runner's thread re-participate in scheduling.
+    ///
+    /// It's used in the scenario of the scale workers. It will treat this
+    /// thread as a new thread. It's guaranteed to be called after `start`,
+    /// before others else.
+    fn reentrant_start(&mut self, _local: &mut Local<Self::TaskCell>) {}
+
     /// Called when a task needs to be handled.
     ///
     /// It's possible that a task can't be finished in a single execution, in
@@ -44,6 +57,16 @@ pub trait Runner {
 
     /// Called when the runner is woken up.
     fn resume(&mut self, _local: &mut Local<Self::TaskCell>) {}
+
+    /// Called when the runner is about to be removed from scheduler.
+    ///
+    /// It's used in the scenario of the scale workers. It will treat this
+    /// thread as a dead thread. It's guaranteed that if this thread is not
+    /// selected by scaling up, no other methods will be called after this
+    /// method, unless the thread pool is shutdown down, in which case `end`
+    /// will be the last method called. Otherwise it's regarded as a new
+    /// thread, will re-run from `reetrant_start`.
+    fn reentrant_end(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
     /// Called when the runner is about to be destroyed.
     ///
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -132,7 +132,7 @@ impl<T> QueueCore<T> {
 
     /// Scale workers.
     pub fn scale_workers(&self, new_thread_count: usize) -> bool {
-        let _ = self.l.lock().unwrap();
+        let _l = self.l.lock().unwrap();
         if new_thread_count < self.config.min_thread_count
             || new_thread_count > self.config.max_thread_count
         {
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -10,7 +10,7 @@ use fail::fail_point;
 use parking_lot_core::{ParkResult, ParkToken, UnparkToken};
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
-    Arc, Weak,
+    Arc, Mutex, Weak,
 };
 
 /// An usize is used to trace the threads that are working actively.
@@ -38,7 +38,11 @@ pub fn is_shutdown(cnt: usize) -> bool {
 pub(crate) struct QueueCore<T> {
     global_queue: TaskInjector<T>,
     active_workers: AtomicUsize,
+    core_workers: AtomicUsize,
     config: SchedConfig,
+    // Only used to protect the scale_workers method to avoid multiple threads
+    // from adjusting the thread pool state at the same time
+    l: Mutex<()>,
 }
 
 impl<T> QueueCore<T> {
@@ -46,7 +50,9 @@ impl<T> QueueCore<T> {
         QueueCore {
             global_queue,
             active_workers: AtomicUsize::new(config.max_thread_count << WORKER_COUNT_SHIFT),
+            core_workers: AtomicUsize::new(config.max_thread_count),
             config,
+            l: Mutex::new(()),
         }
     }
 
@@ -56,7 +62,9 @@ impl<T> QueueCore<T> {
     /// the action.
     pub fn ensure_workers(&self, source: usize) {
         let cnt = self.active_workers.load(Ordering::SeqCst);
-        if (cnt >> WORKER_COUNT_SHIFT) >= self.config.max_thread_count || is_shutdown(cnt) {
+        if (cnt >> WORKER_COUNT_SHIFT) >= self.config.core_thread_count.load(Ordering::SeqCst)
+            || is_shutdown(cnt)
+        {
             return;
         }
 
@@ -74,6 +82,7 @@ impl<T> QueueCore<T> {
         let addr = self as *const QueueCore<T> as usize;
         unsafe {
             parking_lot_core::unpark_all(addr, UnparkToken(source));
+            parking_lot_core::unpark_all(addr | 1, UnparkToken(source));
         }
     }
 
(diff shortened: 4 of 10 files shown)


ethercflow:
@BusyJay @sticnarf PTAL, thanks!

src/pool.rs
```diff
@@ -34,6 +34,14 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
+    /// Scale workers of the thread pool, the new thread count should be
+    /// smaller than the max_thread_count.
```
sticnarf: Better to describe the meaning of the return value.

ethercflow:
I found that the CI failed caused by ```thread 'pool::tests::test_basic' panicked at 'called `Result::unwrap()` on an `Err` value: Timeout', src/pool/tests.rs:38:67,``` but I cannot reproduce it locally. Is this just because CI is too slower?

src/pool/spawn.rs
```diff
@@ -120,6 +125,61 @@ impl<T> QueueCore<T> {
             }
         }
     }
+
+    /// Scale workers.
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
```
sticnarf, line 134: Maybe we'd better protect the whole function with a `Mutex`? I'm afraid the status would be strange if this function is called concurrently in multiple threads.

sticnarf:
> I found that the CI failed caused by `` thread 'pool::tests::test_basic' panicked at 'called `Result::unwrap()` on an `Err` value: Timeout', src/pool/tests.rs:38:67, `` but I cannot reproduce it locally. Is this just because CI is too slower?

I don't know. I think the 1 seconds timeout is large enough. I cannot reproduce the failure locally, either.

src/pool/worker.rs
```diff
@@ -39,18 +39,39 @@ where
...
-                Some(t) => t,
-                None => continue,
-            };
-            self.runner.handle(&mut self.local, task.task_cell);
+        if !self.local.runnable() {
+            self.local.sleep();
         }
-        self.runner.end(&mut self.local);
+        loop {
+            self.runner.start(&mut self.local);
```
sticnarf: Initially, `start` and `end` are only called once. Maybe it is not appropriate to change the semantics. `pause` and `resume` are more suitable here.

src/queue.rs
```diff
@@ -93,10 +93,10 @@ impl<T: TaskCell + Send> LocalQueue<T> {
 
     /// Gets a task cell from the queue. Returns `None` if there is no task cell
     /// available.
-    pub fn pop(&mut self) -> Option<Pop<T>> {
+    pub fn pop(&mut self, need_retry: bool) -> Option<Pop<T>> {
```
sticnarf: I think `need_steal` is a more accurate name

sticnarf (commented):
The idea generally looks good. Is there any benchmark result about this change? It looks it adds a few checks to paths that are frequently called.

src/pool.rs
ethercflow: Agree, will be added

src/pool/worker.rs
ethercflow: Do `start` and `end` represent the life cycle of a runner? What I expect is that executing `start` again means that it has nothing to do with the previous runner, which is equivalent to a new thread?

src/queue.rs
ethercflow: Oh indeed

ethercflow:
> The idea generally looks good. Is there any benchmark result about this change? It looks it adds a few checks to paths that are frequently called.

Not yet, will follow up

src/pool/worker.rs
sticnarf: `Runner::run` is the only thing a worker thread executes. So the lifecycle of a runner equals to the lifecycle of a thread. Initializations or teardowns related to a thread are in the `start` and `end` method, so they should be called only once.

ethercflow: So I think my modification is correct. When a thread finds that it is no longer `runnable`, it means that it needs to be destroyed, but we don’t really destroy it, we just call `end` then park it to another queue (the key is not `QueueCore<T>`'s addr, the key is `QueueCore<T>`'addr | 1), no longer participate in the scheduling of `Remote`. When it is `runnable` again, it already represents a new thread, so it needs to call `start` again.

sticnarf: OK, I've got what you mean. But it still makes me feel a bit uncomfortable that the thread is just "pretended to be" destroyed.

It may be surprising to users that the `start` method is called multiple times for a thread. It may cause some problems if it  sets something in the thread-local storage but not clean it in the `end` method, or it depends on that the thread id from the `start` method cannot duplicate.

Maybe a safe approach is to define two more trait methods in the `Runner` trait for the scaled threads.

@BusyJay What do you think?

ethercflow:
> > The idea generally looks good. Is there any benchmark result about this change? It looks it adds a few checks to paths that are frequently called.
> 
> Not yet, will follow up

The comparison result of bench is as follows, `base` represents the current master branch, and `scale` represents my modification:

image

src/pool/builder.rs
```diff
@@ -5,15 +5,19 @@ use crate::pool::worker::WorkerThread;
...
 use std::thread;
 use std::time::Duration;
 
 /// Configuration for schedule algorithm.
-#[derive(Clone)]
 pub struct SchedConfig {
     /// The maximum number of running threads at the same time.
     pub max_thread_count: usize,
+    /// The core number of running threads at the same time.
+    pub core_thread_count: AtomicUsize,
```
BusyJay, line 23: Can you add some comment to explain why `AtomicUsize` is used?

src/pool/spawn.rs
```diff
@@ -120,6 +125,61 @@ impl<T> QueueCore<T> {
...
+        }
+        loop {
+            match self.core_workers.compare_exchange_weak(
+                core_workers,
+                core_workers - 1,
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            ) {
+                Ok(n) => {
+                    if n > core_thread_count {
```
BusyJay, line 178: What if `core_thread_count` is updated after the loading at L160?

src/pool/worker.rs
BusyJay: Agree with @sticnarf.

src/pool.rs
ethercflow: Fixed, thanks!

src/pool/builder.rs
ethercflow: Fixed, thanks!

src/pool/spawn.rs
ethercflow: It has been added, but I think it is better to be controlled by a management thread in actual use. Do we need annotations?

src/pool/spawn.rs
ethercflow: This path will only be reached when in scaling down.
- If the modification is also due to scale down, then it does not matter. The next time this function is executed by some thread, it will be re-judged with latest value, and do scale down.
- If the modification is due to scale up but `self.core_workers` hasn't been updated by `scale_workers`, eg: it's doing loop `unpack_one`, then it doesn't matter too.
-  If the modification is due to scale up but `self.core_workers` has been updated by `scale_workers` too and before `self.core_workers.compare_exchange_weak`,  the value of `core_workers` may be incorrectly reduced, re-execute this scale up operation can be repaired. But I think the probability of triggering this problem in actual scenes is quite small, because scale itself should be a very low frequency event and also needs to be treated with caution.

src/queue.rs
ethercflow: Fixed, thanks!

src/pool/worker.rs
ethercflow: Add two method `reentrant_start` and `reetrant_end`, does it make sense?

src/pool/spawn.rs
```diff
@@ -142,11 +208,17 @@ impl<T: TaskCell + Send> QueueCore<T> {
 /// dropped.
 pub struct Remote<T> {
     pub(crate) core: Arc<QueueCore<T>>,
+    // Only used to protect the scale_workers method to avoid multiple threads
```
sticnarf: Can we define the `Mutex` in the core? So we don't need an extra `Arc`.

src/pool/spawn.rs
sticnarf: It is better to guarantee thread safety of public functions if possible.

Can you please explain the reason why it should be controlled by a management thread? I think we can assume this function won't be called frequently, so such a `Mutex` will not cause performance issues in most cases.

ethercflow: Oh, I thought that using multiple threads to scale threads might feel a bit weird. But I agree with you,  as much as possible to ensure that the pub interface is thread-safe. I'll modify `Mutex`'s place as your another advice. Thanks!

src/pool/spawn.rs
ethercflow: Fixed, thanks!

src/pool/builder.rs
```diff
@@ -134,6 +157,16 @@ impl Builder {
         self
     }
 
+    /// Sets the core number of running threads at the same time.
```
BusyJay: It's confusing what is `core`.

BusyJay: It should be stated what happen if the count is larger than the `max_thread_count`.

src/pool/runner.rs
```diff
@@ -45,6 +52,14 @@ pub trait Runner {
     /// Called when the runner is woken up.
     fn resume(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
+    /// Called when the runner is about to be removed from scheduler.
+    ///
+    /// It's used in the scenario of the scale workers. It will treat this
+    /// thread as a dead thread. It's guaranteed that no other method will
+    /// be called after this method, unless the thread pool is shutdown down,
+    /// in which case end will be the last mtehod called.
```
BusyJay: `end` should be quoted.

> unless the thread pool is shutdown down

I think other methods will be called again when being scaled up.

src/pool/runner.rs
```diff
@@ -30,6 +30,13 @@ pub trait Runner {
     /// It's guaranteed to be the first method to call before anything else.
     fn start(&mut self, _local: &mut Local<Self::TaskCell>) {}
 
+    /// Called when the runner's thread re-participate in scheduling.
+    ///
+    /// It's used in the scenario of the scale workers. It will treat this
+    /// thread as a new thread. It's guaranteed to be called after start,
+    /// before others else.
+    fn reentrant_start(&mut self, _local: &mut Local<Self::TaskCell>) {}
```
BusyJay, line 44: Can you update the ASCII graph in the trait definition to show the lifecycle?

ethercflow: Fixed, thanks!

src/pool/builder.rs
ethercflow: Yes, the description of selecting and unselecting during scaling up has been corrected, thank you.

src/pool/runner.rs
ethercflow: Yes, the description of selecting and unselecting during scaling up has been corrected, thank you.

src/pool/builder.rs
ethercflow: Yes, but I did it in the freeze_with_queue function. Because we did not limit the order of function calling for setting these three thread sizes.

BusyJay: Yes, but it should still be stated.

ethercflow: Oh, I misunderstood what you mean, fixed thanks!

ethercflow: Made some adjustments, does it make sense now?

ethercflow:
> > I found that the CI failed caused by `` thread 'pool::tests::test_basic' panicked at 'called `Result::unwrap()` on an `Err` value: Timeout', src/pool/tests.rs:38:67, `` but I cannot reproduce it locally. Is this just because CI is too slower?
> 
> I don't know. I think the 1 seconds timeout is large enough. I cannot reproduce the failure locally, either.

I found the reason. The setting of the default value of `core_thread_count` is problematic. If it is not set, it should be equal to the `max_thread_count` not `cpu_nr`.

`taskset -a -c 1 cargo bench --all - --test` can reproduce.

src/pool/spawn.rs
```diff
@@ -120,6 +129,68 @@ impl<T> QueueCore<T> {
             }
         }
     }
+
+    /// Scale workers.
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
+        let _ = self.l.lock().unwrap();
```
BusyJay: The lock is released immediately as the returning value is dropped.

src/pool/spawn.rs
```diff
@@ -120,6 +129,68 @@ impl<T> QueueCore<T> {
...
+            return false;
+        }
+
+        let current_thread_count = self
+            .config
+            .core_thread_count
+            .swap(new_thread_count, Ordering::SeqCst);
+        let gap = new_thread_count as isize - current_thread_count as isize;
+        if gap > 0 {
+            let addr = self as *const QueueCore<T> as usize | 1;
```
BusyJay: Why not `+ 1`?

src/pool/worker.rs
```diff
@@ -39,27 +39,52 @@ where
...
+                self.runner.handle(&mut self.local, task.task_cell);
+                if !self.local.runnable() {
+                    break;
+                }
+            }
+            self.runner.reentrant_end(&mut self.local);
+
+            // If this worker should go to sleep, spawn all futures to remote
+            while let Some(t) = self.local.pop(false) {
+                if self.local.core().is_shutdown() {
```
BusyJay: Should be checked before the loop.

src/pool/spawn.rs
BusyJay: > actual scenes is quite small

Can it be made correct?

ethercflow: I did not think of a solution before, because two atomic variables have to be manipulated

src/pool/spawn.rs
ethercflow: I think both are okay? What are the benefits of choosing  `+1`?  I remember this value is used as a number.

src/pool/worker.rs
ethercflow: I think it should be checked in a loop, because this loop is mainly to spawn unprocessed tasks from the local queue to the global queue during scale down. If shutdown occurs during this process, we stop the spawn process and discard it directly.

Otherwise, I think there will be an endless loop, local will spawn tasks to the global, and then continue to steal all tasks from the global to drain all tasks.

ethercflow: There is a check below this loop, and that check is for “pure shutdown scenarios”

src/pool/spawn.rs
ethercflow: Oh, fixed thanks!

src/pool/spawn.rs
BusyJay: `| 1` assumes address always be even.

src/pool/worker.rs
BusyJay: > I think there will be an endless loop,

Given that steal is false, how can it be an endless loop?

ethercflow: No steal is needed when scale down, but steal is needed when shut down.

src/pool/spawn.rs
ethercflow: Oh, you are right, it will be fixed, thanks!

src/pool/worker.rs
ethercflow: >> I think there will be an endless loop,

>Given that steal is false, how can it be an endless loop?

I was wrong, there will be no endless loop. It's just that there will be some meaningless spawn_remote operations.

src/pool/spawn.rs
glorv: How about change `core_workers` by each thread at reentrant_start/end, thus its count can always represent the exact number of active threads.

src/pool/spawn.rs
```diff
@@ -120,6 +129,68 @@ impl<T> QueueCore<T> {
...
+            ) {
+                Ok(n) => {
+                    if n > core_thread_count {
+                        return true;
+                    } else {
+                        return false;
+                    }
+                }
+                Err(n) => {
+                    if n == core_thread_count {
```
glorv, line 185:
```suggestion
                    if n <= core_thread_count {
```
`core_thread_count` might be scaled down, and now it is a outdated value.

PR #65, opened 2022-02-14
Support dynamically scaling workers of the pool

ethercflow:
Simplified implementation of [previous version](https://github.com/tikv/yatp/pull/61):

1. Unified use of one Queue to manage all threads;
2. Use `active_workers` to consider if it need to scale up as before;
3. Use `unpack_filter` instead of `unpack_one` to determine which thread to unpark by comparing `ParkToken(id)` and `core_thread_count`, only thread id not larger than this value can be unparked;
4. Restore runner and worker implementation;

Test scale with [test-yatp](https://github.com/ethercflow/test-yatp) as before.

The test logic is:

``` rust
fn scale_workers() {
    let pool = Builder::new("SP")
        .max_thread_count(16)
        .core_thread_count(4)
        .build_callback_pool();
    let handler = pool.remote().clone();
    let builder = thread::Builder::new().name("wl".to_string());
    builder
        .spawn(move || {
            loop {
                let (tx, rx) = mpsc::channel();
                // A bunch of tasks should be executed correctly.
                let cases: Vec<_> = (10..100000000).collect();
                for id in &cases {
                    let t = tx.clone();
                    let id = *id;
                    handler.spawn(move |_: &mut Handle<'_>| t.send(id).unwrap());
                }
                let mut ans = vec![];
                for _ in 10..100000000 {
                    let r = rx.recv_timeout(Duration::from_secs(1)).unwrap();
                    ans.push(r);
                }
                ans.sort();
                assert_eq!(cases, ans);
                println!("finish one loop");
            }
        })
        .unwrap();
    loop {
        let mut rng = rand::thread_rng();
        let new_thread_count = rng.gen_range(1..16);
        println!("scale workers to {}", new_thread_count);
        pool.remote().scale_workers(new_thread_count);
        thread::sleep(Duration::from_secs(10));
    }
}
```

While running, execute the `pidstat -G yatp -t 1` command to observe that the number of threads running is in line with expectations:

``` sh
➜  test-yatp git:(main) ./target/release/test-yatp
scale workers to 8
scale workers to 4
scale workers to 9
```

``` sh
00:57:19      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:20        0   3950591         -  785.00   77.00    0.00    0.00  862.00    17  test-yatp
00:57:20        0         -   3950592   89.00    6.00    0.00    0.00   95.00     6  |__SP-0
00:57:20        0         -   3950593   87.00    8.00    0.00    0.00   95.00     2  |__SP-1
00:57:20        0         -   3950594   90.00    6.00    0.00    0.00   96.00     7  |__SP-2
00:57:20        0         -   3950595   87.00    9.00    0.00    0.00   96.00    10  |__SP-3
00:57:20        0         -   3950604   88.00    8.00    0.00    0.00   96.00    39  |__SP-4
00:57:20        0         -   3950605   88.00    8.00    0.00    0.00   96.00    31  |__SP-5
00:57:20        0         -   3950606   87.00    8.00    0.00    0.00   95.00    33  |__SP-6
00:57:20        0         -   3950607   87.00    8.00    0.00    0.00   95.00    21  |__SP-7
00:57:20        0         -   3950608   83.00   16.00    0.00    0.00   99.00    25  |__wl

00:57:20      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:21        0   3950591         -  797.00  149.00    0.00    0.00  946.00    17  test-yatp
00:57:21        0         -   3950592   75.00   13.00    0.00    0.00   88.00    21  |__SP-0
00:57:21        0         -   3950593   76.00   10.00    0.00    0.00   86.00     7  |__SP-1
00:57:21        0         -   3950594   75.00   11.00    0.00    1.00   86.00    31  |__SP-2
00:57:21        0         -   3950595   75.00   11.00    0.00    0.00   86.00    10  |__SP-3
00:57:21        0         -   3950608   69.00   32.00    0.00    0.00  101.00    25  |__wl

00:57:34      UID      TGID       TID    %usr %system  %guest   %wait    %CPU   CPU  Command
00:57:35        0   3950591         -  829.00  100.00    0.00    0.00  929.00    21  test-yatp
00:57:35        0         -   3950592   84.00    8.00    0.00    0.00   92.00    30  |__SP-0
00:57:35        0         -   3950594   83.00   10.00    0.00    0.00   93.00    15  |__SP-1
00:57:35        0         -   3950595   85.00    8.00    0.00    0.00   93.00    17  |__SP-2
00:57:35        0         -   3950597   83.00    9.00    0.00    1.00   92.00     0  |__SP-3
00:57:35        0         -   3950599   85.00    7.00    0.00    0.00   92.00    29  |__SP-4
00:57:35        0         -   3950600   82.00   10.00    0.00    0.00   92.00    34  |__SP-5
00:57:35        0         -   3950602   83.00    9.00    0.00    0.00   92.00    25  |__SP-6
00:57:35        0         -   3950606   85.00    8.00    0.00    0.00   93.00    23  |__SP-7
00:57:35        0         -   3950607   82.00   10.00    0.00    1.00   92.00     1  |__SP-8
00:57:35        0         -   3950608   78.00   19.00    0.00    0.00   97.00    31  |__wl
^C
```

The bench cmp result is here:

image

Signed-off-by: Wenbo Zhang <ethercflow@gmail.com>

### DIFF
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -34,22 +34,24 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
-    /// Scale workers of the thread pool, the new thread count should be less
-    /// than or equal to the max_thread_count, otherwise it will become no-op
-    /// and return false.
-    ///
-    /// If the pool is shutdown, it becomes no-op.
+    /// Scale workers of the thread pool, the adjustable range is `min_thread_count`
+    /// to `max_thread_count`, if this value exceeds `max_thread_count` or zero, it
+    /// will be adjusted to `max_thread_count`, if this value is between zero and
+    /// `min_thread_count`, it will be adjusted to `min_thread_count`.
     ///
     /// Notice:
     /// 1. The effect of scaling may be delayed, eg:
-    ///    - Threads run long-term tasks, resulting in inability to scale down in time,
-    ///      until they have no task to process and can sleep;
-    ///    - Scaling up relies on spawning tasks to the thread pool to unpark new threads,
-    ///      so the delay depends on the time interval between scale and spawn;
-    /// 2. If lots of tasks have been spawned before start, all threads (max_thread_count)
-    ///    will run until they can sleep, then scheduled based on core_thread_count;
-    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
-        self.remote.scale_workers(new_thread_count)
+    ///    - Threads run long-term tasks, resulting in inability to scale down in
+    ///      time, until they have no task to process and can sleep;
+    ///    - Scaling up relies on spawning tasks to the thread pool to unpark new
+    ///      threads, so the delay depends on the time interval between scale and
+    ///      spawn;
+    /// 2. If lots of tasks have been spawned before start, there may be more than
+    ///    `core_thread_count` (max up to `max_thread_count`) threads running at a
+    ///    moment until they can sleep (no tasks to run), then scheduled based on
+    ///    `core_thread_count`;
+    pub fn scale_workers(&self, new_thread_count: usize) {
+        self.remote.scale_workers(new_thread_count);
     }
 
     /// Shutdowns the pool.
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -34,11 +34,20 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
-    /// Scale workers of the thread pool, the new thread count should be
-    /// smaller than the max_thread_count, otherwise it will become no-op
+    /// Scale workers of the thread pool, the new thread count should be less
+    /// than or equal to the max_thread_count, otherwise it will become no-op
     /// and return false.
     ///
     /// If the pool is shutdown, it becomes no-op.
+    ///
+    /// Notice:
+    /// 1. The effect of scaling may be delayed, eg:
+    ///    - Threads run long-term tasks, resulting in inability to scale down in time,
+    ///      until they have no task to process and can sleep;
+    ///    - Scaling up relies on spawning tasks to the thread pool to unpark new threads,
+    ///      so the delay depends on the time interval between scale and spawn;
+    /// 2. If lots of tasks have been spawned before start, all threads (max_thread_count)
+    ///    will run until they can sleep, then scheduled based on core_thread_count;
     pub fn scale_workers(&self, new_thread_count: usize) -> bool {
         self.remote.scale_workers(new_thread_count)
     }
--- a/src/pool.rs
+++ b/src/pool.rs
@@ -34,6 +34,15 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
+    /// Scale workers of the thread pool, the new thread count should be
+    /// smaller than the max_thread_count, otherwise it will become no-op
+    /// and return false.
+    ///
+    /// If the pool is shutdown, it becomes no-op.
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
+        self.remote.scale_workers(new_thread_count)
+    }
+
     /// Shutdowns the pool.
     ///
     /// Closes the queue and wait for all threads to exit.
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -138,16 +138,15 @@ impl<T> QueueCore<T> {
     }
 
     /// Scale workers.
-    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
-        if new_thread_count < self.config.min_thread_count
-            || new_thread_count > self.config.max_thread_count
-        {
-            return false;
+    pub fn scale_workers(&self, mut new_thread_count: usize) {
+        if new_thread_count == 0 || new_thread_count > self.config.max_thread_count {
+            new_thread_count = self.config.max_thread_count;
+        } else if new_thread_count < self.config.min_thread_count {
+            new_thread_count = self.config.min_thread_count;
         }
         self.config
             .core_thread_count
             .store(new_thread_count, Ordering::SeqCst);
-        true
     }
 }
 
@@ -185,7 +184,7 @@ impl<T: TaskCell + Send> Remote<T> {
     }
 
     /// Scales workers of the thread pool.
-    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
+    pub fn scale_workers(&self, new_thread_count: usize) {
         self.core.scale_workers(new_thread_count)
     }
 
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -7,7 +7,7 @@
 use crate::pool::SchedConfig;
 use crate::queue::{Extras, LocalQueue, Pop, TaskCell, TaskInjector, WithExtras};
 use fail::fail_point;
-use parking_lot_core::{ParkResult, ParkToken, UnparkToken};
+use parking_lot_core::{FilterOp, ParkResult, ParkToken, UnparkToken};
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
     Arc, Weak,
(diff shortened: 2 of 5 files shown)


src/pool.rs
```diff
@@ -34,6 +34,15 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
+    /// Scale workers of the thread pool, the new thread count should be
+    /// smaller than the max_thread_count, otherwise it will become no-op
```
BusyJay: Perhaps should be adjusted to `max_thread_count` to keep consistent with `Builder`.

src/pool/spawn.rs
```diff
@@ -56,13 +56,29 @@ impl<T> QueueCore<T> {
...
 
         let addr = self as *const QueueCore<T> as usize;
+        let mut unparked_once = false;
+
         unsafe {
-            parking_lot_core::unpark_one(addr, |_| UnparkToken(source));
+            parking_lot_core::unpark_filter(
+                addr,
+                |p: ParkToken| {
+                    if !unparked_once && p.0 <= self.config.core_thread_count.load(Ordering::SeqCst)
```
BusyJay, line 72: If it's scheduled a lot of tasks before start, then all threads will have workload despite the `core_thread_count`, isn't it?

BusyJay (commented):
Much clean! How does it perform?

src/pool/spawn.rs
```diff
@@ -120,6 +136,19 @@ impl<T> QueueCore<T> {
...
+    /// Scale workers.
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
+        if new_thread_count < self.config.min_thread_count
+            || new_thread_count > self.config.max_thread_count
+        {
+            return false;
+        }
+        self.config
+            .core_thread_count
+            .store(new_thread_count, Ordering::SeqCst);
```
BusyJay, line 149: If all threads are occupied by long running tasks, then scale workers may take affect after a long time. It maybe a surprise if people try to use it to shorten pending queue length.

ethercflow:
Bench result：

image

image

image

image

image

image

image

image

image

image

image

image

image

src/pool/spawn.rs
glorv: I think this is expected behavior if we want to ensure scale is gracefully handled.

src/pool/spawn.rs
```diff
@@ -56,13 +56,29 @@ impl<T> QueueCore<T> {
...
-            parking_lot_core::unpark_one(addr, |_| UnparkToken(source));
+            parking_lot_core::unpark_filter(
+                addr,
+                |p: ParkToken| {
+                    if !unparked_once && p.0 <= self.config.core_thread_count.load(Ordering::SeqCst)
+                    {
+                        unparked_once = true;
+                        FilterOp::Unpark
+                    } else {
+                        FilterOp::Skip
```
BusyJay, line 77: Why not `FilterOp::Stop`?

src/pool.rs
```diff
@@ -34,6 +34,24 @@ impl<T: TaskCell + Send> ThreadPool<T> {
         self.remote.spawn(t);
     }
 
+    /// Scale workers of the thread pool, the new thread count should be less
+    /// than or equal to the max_thread_count, otherwise it will become no-op
```
BusyJay: It should follow the same behavior of `Builder::core_thread_count`.

src/pool.rs
```diff
@@ -34,6 +34,24 @@ impl<T: TaskCell + Send> ThreadPool<T> {
...
+    ///
+    /// Notice:
+    /// 1. The effect of scaling may be delayed, eg:
+    ///    - Threads run long-term tasks, resulting in inability to scale down in time,
+    ///      until they have no task to process and can sleep;
+    ///    - Scaling up relies on spawning tasks to the thread pool to unpark new threads,
+    ///      so the delay depends on the time interval between scale and spawn;
+    /// 2. If lots of tasks have been spawned before start, all threads (max_thread_count)
+    ///    will run until they can sleep, then scheduled based on core_thread_count;
+    pub fn scale_workers(&self, new_thread_count: usize) -> bool {
```
BusyJay: Any tests?

src/pool/spawn.rs
ethercflow: Because if stop, the next execution will start from the position that caused stop, which will be problematic, for example:

core = 4, max = 16
queue is 1, 11, 2, 3

The first ensure_worker will wakeup 1 and stop at 11, and subsequent ensure_worker will always stop at 11.

src/pool.rs
ethercflow: I only did system testing, I will think about adding unit tests.

src/pool.rs
ethercflow: Sure

ethercflow: Fixed (Sorry, I misunderstood what you meant by the same comment last time)

src/pool.rs
ethercflow: > I only did system testing, I will think about adding unit tests.

I had done three tests:
1. Spawn 10000 tasks first, then lazy build threads, from `pidstat -G yatp -t 1`, I can see in an instant `max_thread_count` threads are woken up to work and then drop to `core_thread_count` (I continue to spawn tasks);
2. Call scale_workers with `0`, `< min_thread_count`, `> max_thread_count`, I can see it will adjust to `max_thread_count`, `min_thread_count`, `max_thread_count`;
3. Continuous spawn task, call `scale_workers` every 3s, the parameter is a random number between 0~`max_thread_count`, observed `pidstat -G yatp -t 1` equal to scaled  thread number;

Do you think these are enough? Also, I don't think these test cases fit in tests.rs . eg: not easy to get the number of running threads.

BusyJay: You can use blocking channels to test if all threads are woken up and block in unit tests.

ethercflow: > You can use blocking channels to test if all threads are woken up and block in unit tests.

Added two test cases, thanks!

ethercflow:
See a large performance regression from the last bench result, re-run five times to determine if it is a disturbance
image

The new results:
image

From the results, I think there is a disturbance in the last test, not a real performance regression.

src/pool/tests.rs
```diff
@@ -167,3 +167,130 @@ fn test_shutdown_with_futures() {
...
+    pool.spawn(move |_: &mut Handle<'_>| tx.send(1).unwrap());
+    // No runnable threads, so it should time out
+    assert_eq!(
+        rx.recv_timeout(Duration::from_secs(1)),
+        Err(mpsc::RecvTimeoutError::Timeout)
+    );
+
+    // Scale down two workers
+    pool.scale_workers(1);
+    // Due to the current implementation, the scale down cann't be triggered until thread
```
BusyJay:
```suggestion
    // Due to the current implementation, the scale down can't be triggered until thread
```

src/pool/tests.rs
```diff
@@ -167,3 +167,130 @@ fn test_shutdown_with_futures() {
...
+    let tx1 = tx.clone();
+    pool.spawn(move |_: &mut Handle<'_>| tx1.send(1).unwrap());
+    // No runnable threads, so it should time out
+    assert_eq!(
+        rx.recv_timeout(Duration::from_secs(1)),
+        Err(mpsc::RecvTimeoutError::Timeout)
+    );
+    // Scale up one worker
+    pool.scale_workers(3);
+    // Due to the current implementation, the scale up cann't be triggered until a new
```
BusyJay:
```suggestion
    // Due to the current implementation, the scale up can't be triggered until a new
```

BusyJay (dismissed):

ethercflow:
The last force-push corrected the misspelling of the comment word (because after I directly accepted your modification, github prompted the lack of sign-off information)

src/pool/spawn.rs
sticnarf: So, this means the `unpark` is an `O(n)` operation where `n` is the number of parked threads?

sticnarf: Do you have benchmark result when `core_thread_count` is much smaller than `max_thread_count` (e.g. 4 out of 32)

ethercflow: > Do you have benchmark result when `core_thread_count` is much smaller than `max_thread_count` (e.g. 4 out of 32)

I'll upload it later, thanks!

ethercflow:
> Do you have benchmark result when core_thread_count is much smaller than max_thread_count (e.g. 4 out of 32)

@sticnarf Here's the result, PTAL.

image
image
image
image

I chose to re-test with the multilevel/512 a few times, and it looks like that there was no significant performance impact.

image

sticnarf (approved):
LGTM

PR #67, opened 2022-08-12
add some new metrics

glorv:
This PR add following new metrics:
- TASK_WAIT_DURATION. Task total wait duration in queue.
- TASK_EXEC_DURATION. Task total execution duration.
- TASK_POLL_DURATION. Tack execution duration during one single pull.
- TASK_EXEC_TIMES. Task total execution/schedule times.

Currently, these metreics is only added for multi-level queue, we can also support for single level queue if needed.

### DIFF
--- a/src/queue/extras.rs
+++ b/src/queue/extras.rs
@@ -9,19 +9,25 @@ use std::time::{Duration, Instant};
 /// The extras for the task cells pushed into a queue.
 #[derive(Debug, Clone)]
 pub struct Extras {
+    pub(crate) start_time: Instant,
     /// The instant when the task cell is pushed to the queue.
     pub(crate) schedule_time: Option<Instant>,
     /// The identifier of the task. Only used in the multilevel task queue.
     pub(crate) task_id: u64,
-    /// The time spent on handling this task. Only used in the multilevel task
-    /// queue.
+    // The total time spent on handling this task.
     pub(crate) running_time: Option<Arc<ElapsedTime>>,
+    /// The total time spent on handling this task. Only used in the multilevel task
+    /// queue. This total time will accumulate the execution time acorss multiple tasks
+    /// with the same task_id.
+    pub(crate) total_running_time: Option<Arc<ElapsedTime>>,
     /// The level of queue which this task comes from. Only used in the
     /// multilevel task queue.
     pub(crate) current_level: u8,
     /// If `fixed_level` is `Some`, this task is always pushed to the given
     /// level. Only used in the multilevel task queue.
     pub(crate) fixed_level: Option<u8>,
+    /// Number of execute times
+    pub(crate) exec_times: u32,
 }
 
 impl Extras {
@@ -29,11 +35,14 @@ impl Extras {
     /// task queue.
     pub fn single_level() -> Extras {
         Extras {
+            start_time: Instant::now(),
             schedule_time: None,
             task_id: 0,
             running_time: None,
+            total_running_time: None,
             current_level: 0,
             fixed_level: None,
+            exec_times: 0,
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/queue/multilevel.rs
```diff
@@ -315,6 +362,7 @@ impl LevelManager {
         };
         extras.current_level = current_level;
         extras.schedule_time = Some(now());
+        self.tasks_per_level[current_level as usize].inc();
```
sticnarf: This is a global atomic value that is mutated very frequently. I'm thinking whether this can be avoided by making this a [Collector](https://docs.rs/prometheus/latest/prometheus/core/trait.Collector.html#). It can be a wrapper of a normal `Gauge` but we can update it lazily.

So, only when Prometheus pulls the data or when we want to use this to adjust some internal parameters, we call `Injector::len` to get the length to avoid extra atomic operations.

src/queue/multilevel.rs
```diff
@@ -274,6 +311,11 @@ where
     }
 }
 
+fn duration_to_sec(d: Duration) -> f64 {
```
sticnarf: What about just using `Duration::to_secs_f64`

src/queue/extras.rs
```diff
@@ -9,31 +9,40 @@ use std::time::{Duration, Instant};
...
 pub struct Extras {
+    pub(crate) start_time: Instant,
     /// The instant when the task cell is pushed to the queue.
     pub(crate) schedule_time: Option<Instant>,
     /// The identifier of the task. Only used in the multilevel task queue.
     pub(crate) task_id: u64,
-    /// The time spent on handling this task. Only used in the multilevel task
-    /// queue.
+    // The total time spent on handling this task.
     pub(crate) running_time: Option<Arc<ElapsedTime>>,
```
sticnarf, line 18: It seems to me that this won't be shared across threads... So, an `Arc<ElapsedTime>` seems unnecessary?

src/queue/multilevel.rs
```diff
@@ -211,6 +211,15 @@ where
             manager: self.manager.clone(),
             local_level0_elapsed_us: self.manager.level0_elapsed_us.local(),
             local_total_elapsed_us: self.manager.total_elapsed_us.local(),
+            tasks_per_level: self.manager.tasks_per_level.clone(),
+            task_execute_duration: self.manager.task_execute_duration.local(),
+            task_wait_duration: self.manager.task_wait_duration.local(),
+            task_poll_duration: self
```
sticnarf: To avoid nightly features, I think you can just write it as `array::from_fn(|i| self.manager.task_poll_duration[i].local())`

src/queue/multilevel.rs
glorv: Yeah, use a Collector is a good idea here. Changing to collecor need a global lazy static variable to store a weak ref of all running injectors, this need to change some of the structs. But I found we already had `tikv_unified_read_pool_running_tasks`, so no need to add another similar metrics. I'm going to remove this metric anyway.

src/queue/extras.rs
glorv: This is necessary because the task is moved when execute the task in worker inner

sticnarf: Yes, it will be moved across thread boundaries. But it isn't modified concurrently by different threads, is it?

glorv: No, it won't be visited by different thread concurrency. Use arc just to make the compiler happy. 😅

sticnarf: Ah, I see. And I find it actually possible to be modified concurrently. After the inner future runner finishes polling, the task may be picked by another thread. At this time, there can be more than one threads in the `handle` method for the same task.

sticnarf (dismissed):
LGTM. But you have to fix the DCO check.

sticnarf (approved):

PR #68, opened 2022-08-16
simplify the implement of mark_sleep and mark_woken

glorv:


src/pool/spawn.rs
```diff
@@ -103,38 +103,22 @@ impl<T> QueueCore<T> {
     ///
     /// It can be marked as sleep only when the pool is not shutting down.
     pub fn mark_sleep(&self) -> bool {
-        let mut cnt = self.active_workers.load(Ordering::SeqCst);
-        loop {
-            if is_shutdown(cnt) {
```
sticnarf, line 108: Yep, the active count should be not important once the pool is shut down.

@BusyJay Is there any special reason to check shutdown first?

sticnarf (approved):

src/pool/spawn.rs
BusyJay: No. Combining a loop and `compare_exchange_weak` just to be more performant when there is something to check.

PR #69, opened 2022-08-24
use different batch size for each level queue

glorv:
This is an attempt to make the schedule fairer. In some benchmark, I found some slow queries can cause huge performance issue to the overall performance both in throughput and latency. This kind of slow queries often spawn a lot of copr tasks and many of them are quite costly. One cause of this performance regression is that tasks from high level queue are spawned more times than expected that slow down the overall performance.

As in the current implement, task are stealed in batch, the batch-size should also consider the execution percentage.
In my benchmark, I found even in some heavy load(e.g. sysbench oltp_readonly with 512 threads), the avg batch-size of one steal is small than 10 (P50 = 1.2, P90=10.5), thus the batch-size for level1 should less than 2, while level2 less than 1. To make things simple, we can just only steal one task from level1 and level2 at each time.  

There is one drawback that this strategy change can make low priority tasks(or long-running tasks) executing slower because their execution possibility is decreased.

An alternatiy is to add a method `steal_batch_by_and_pop(usize)` to explicitly set the task count that need to be stealed, and set different value for each level.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,9 @@ license = "Apache-2.0"
 repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
-crossbeam-deque = "0.8"
+# TODO: switch back to stable version when the next version is released.
+crossbeam-deque = {git = "https://github.com/crossbeam-rs/crossbeam", rev="41ed3d948720f26149b2ebeaf58fe8a193134056"}
+#crossbeam-deque = "0.8"
 dashmap = "5.1"
 fail = "0.5"
 lazy_static = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,9 +9,7 @@ license = "Apache-2.0"
 repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
-# TODO: switch back to stable version when the next version is released.
-crossbeam-deque = {git = "https://github.com/crossbeam-rs/crossbeam", rev="41ed3d948720f26149b2ebeaf58fe8a193134056"}
-#crossbeam-deque = "0.8"
+crossbeam-deque = "0.8"
 dashmap = "5.1"
 fail = "0.5"
 lazy_static = "1"
@@ -19,6 +17,10 @@ num_cpus = "1.0"
 parking_lot_core = "0.9"
 prometheus = { version = "0.13", default-features = false }
 rand = "0.8"
+
+# TODO: remove this patch after the next version is released.
+[patch.crates-io]
+crossbeam-deque = {git = "https://github.com/crossbeam-rs/crossbeam", rev="41ed3d948720f26149b2ebeaf58fe8a193134056"}
 
 [features]
 failpoints = ["fail/failpoints"]
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -410,6 +410,7 @@ impl LevelManager {
 
         let level_0_tasks = (cur_total_tasks_per_level[0] - last_level0_total_tasks) as usize;
         let total_tasks = (cur_total_tasks - last_total_tasks) as usize;
+        // adjust the batch size after meeting enough tasks.
         if total_tasks > ADJUST_LEVEL_STEAL_SIZE_THRESHOLD {
             let new_steal_count = if level_0_tasks == 0 {
                 // level 0 has no tasks, that means the current workloads are all low-priority tasks.
@@ -422,9 +423,9 @@ impl LevelManager {
             };
             self.max_level_queue_steal_size
                 .store(new_steal_count as usize, SeqCst);
-        }
-        for (i, c) in self.last_exec_tasks_per_level.iter().enumerate() {
-            c.set(cur_total_tasks_per_level[i]);
+            for (i, c) in self.last_exec_tasks_per_level.iter().enumerate() {
+                c.set(cur_total_tasks_per_level[i]);
+            }
         }
 
         self.adjusting.store(false, SeqCst);
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -55,6 +55,13 @@ const MAX_LEVEL0_CHANCE: f64 = 0.98;
 /// The amount that the level 0 chance is increased or decreased each time.
 const ADJUST_AMOUNT: f64 = 0.06;
 
+/// the default max number of tasks the can steal from each level queue.
+const DEFAULT_STEAL_LIMIT_PER_LEVEL: [usize; LEVEL_NUM] = [64, 16, 1];
+/// the maximum number of tasks that should steal from global queue for level > 0 queue.
+const LEVEL_MAX_QUEUE_MAX_STEAL_SIZE: usize = 16;
+/// the number of tasks threshold to trigger adjust level max steal size.
+const ADJUST_LEVEL_STEAL_SIZE_THRESHOLD: usize = 100;
+
 /// The injector of a multilevel task queue.
 pub(crate) struct TaskInjector<T> {
     level_injectors: Arc<[Injector<T>; LEVEL_NUM]>,
@@ -126,7 +133,7 @@ where
                     .find(|_| rng.gen_ratio(CHANCE_RATIO, CHANCE_RATIO + 1))
                     .unwrap_or(LEVEL_NUM - 1)
             };
-            match self.level_injectors[expected_level].steal_batch_and_pop(&self.local_queue) {
+            match self.steal_from_injector(expected_level) {
                 Steal::Success(t) => return Some(into_pop(t, false)),
                 Steal::Retry => need_retry = true,
                 _ => {}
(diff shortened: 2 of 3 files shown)


src/queue/multilevel.rs
```diff
@@ -165,6 +159,20 @@ where
         None
     }
 
+    // steal one task from level injector, the steal strategy is depending on the level:
+    // - level 0. try steal batch to local queue. This can help optimize the performance.
+    // - level >0. Only steal one task once. This is to avoid steal too many low priority task to
```
kevin-xianliu: avoid stealing too many low priority tasks

BusyJay:
Interesting. Do you have any test case or microbench to show the impact?

sticnarf:
Looks good as the prioritized goal is to guarantee the latency of level 0 tasks.

I wonder how much regression it brings if most requests are not in level 0.

glorv:
@BusyJay @sticnarf 
I have do a benchmark based on the workload provided by one of our customers. The goal is to optimize the performance impact of big slow queries. One of their slow queries is that the query spawns more than 500K coprocessor tasks and it cost more than 1min to finished, during it execution, the overall throughput decrease more than 25%.
Some information of the big query:
image
image
image

I use following workload to measure the impact of this kind of big queries: The base workload only runs the basic fast queries with 24 thread; And the test workload add one extra thread that only loop runs the slow query. Each benchmark is only run for 10min as my other longer benchmark show 10min is enough to show the different(the result of 10min and 1h benchmask is relatively near):
| | base test + old code | slow test + old code | base test + new code  | slow test + new code |
| -- | ------ |  ------ | ------ | ------ |
| qps | 4012.47 | 3019.43 | 4112.90 | 3928.58 |
| avg lat | 5.98 | 8.29 | 5.83 | 6.36 |
| p99 lat | 47.47 | 71.83 | 51.94 | 55.82 |
| max lat | 2564.44 | 199951.88 | 1952.12 | 302260.36 |

From the result we can see: when there are no big queries, the performance difference is quite small; when there are big queries(which add many low priority tasks), the overall performance is much better with one drawback that the slow queries is executed much slower.

Some metrics: 
image
image
image

sticnarf:
It's expected and acceptable that big queries run much slower in the mixed workload case. I'm just a bit worried about how much it will slow down when there are only big queries, because now nearly all tasks from big queries are only stolen one by one from the injector.

glorv:
> It's expected and acceptable that big queries run much slower in the mixed workload case. I'm just a bit worried about how much it will slow down when there are only big queries, because now nearly all tasks from big queries are only stolen one by one from the injector.

Ok. I will do another benchmark for only slow queries.

glorv:
I tests workload with only big queries.  Because the big query consumes a lot of resource, I only run the benchmark with 1,2,5,10 threads. As in the most common production case, the concurrency should be only 1.
| | old 1t | new 1t | old 2t | new 2t | old 5t | new 5t | old 10t | new 10t |
| ---- | --- | --- | --- | --- | --- | --- | --- | --- |
| transactions | 29 | 25 | 40 | 32 | 60 | 64 | 103 | 100 |
| avg lat(s) | 21.1 | 24.0 |  31.2 | 39.1 | 52.5 | 51.8 | 69.7 | 72.4 |
| max lat(s) | 56.3 | 53.2 | 70.3 | 83.1 | 108.2 | 117.1 | 153.8 | 158.8 |

NOTE: Because the total number of transactions is relatively small, qps and p99 latency is not showed here.

From this table we can see there are about 15~20% performance loss when the concurrency is small. As this is the common case, I will think about how to handle this regression.

src/queue/multilevel.rs
```diff
@@ -165,6 +159,20 @@ where
...
 
+    // steal one task from level injector, the steal strategy is depending on the level:
+    // - level 0. try steal batch to local queue. This can help optimize the performance.
+    // - level >0. Only steal one task once. This is to avoid stealing too many low priority task to
+    //             local queue that causes level 0 task to wait for too long. In my benchmark, even
+    //             when the load is heavy, the average steal batch size from level 0 is <5, so it fair
+    //             to only pop 1 task from higher level as their execution percentage is <25% of total.
+    fn steal_from_injector(&self, level: usize) -> Steal<T> {
+        if level == 0 {
+            self.level_injectors[level].steal_batch_and_pop(&self.local_queue)
```
kevin-xianliu: We can use a parameter/switch to disable batch-stealing from L1/L2.

kevin-xianliu:
> I tests workload with only big queries. Because the big query consumes a lot of resource, I only run the benchmark with 1,2,5,10 threads. As in the most common production case, the concurrency should be only 1.
> 
> old 1t	new 1t	old 2t	new 2t	old 5t	new 5t	old 10t	new 10t
> transactions	29	25	40	32	60	64	103	100
> avg lat(s)	21.1	24.0	31.2	39.1	52.5	51.8	69.7	72.4
> max lat(s)	56.3	53.2	70.3	83.1	108.2	117.1	153.8	158.8
> NOTE: Because the total number of transactions is relatively small, qps and p99 latency is not showed here.
> 
> From this table we can see there are about 15~20% performance loss when the concurrency is small. As this is the common case, I will think about how to handle this regression.

I think the latency impact is actually acceptable.   Per the results, the max latency difference are so small once the concurrency exceeds 5.

glorv:
@sticnarf I tried to resolve the regression by adjusting the steal number of low priority queue based on the active task numbers. It's much similar to the adjusting of level 0 chance adjust. In my benchmark, after this change, there is no regression in only big queries workload. The change depend on a small change of cross-deque to restrict the task count of each steal and #67 for new metrics. The change is at https://github.com/glorv/yatp/commit/3e724dd1a513fb6d32e4860498294c2aff572b39, what do you think

sticnarf:
> @sticnarf I tried to resolve the regression by adjusting the steal number of low priority queue based on the active task numbers. It's much similar to the adjusting of level 0 chance adjust. In my benchmark, after this change, there is no regression in only big queries workload. The change depend on a small change of cross-deque to restrict the task count of each steal and #67 for new metrics. The change is at [glorv@3e724dd](https://github.com/glorv/yatp/commit/3e724dd1a513fb6d32e4860498294c2aff572b39), what do you think

Yep, looks more complex but optimal.

We can add this API to upstream crossbeam. I think it may be useful to everyone.

glorv:
@sticnarf @BusyJay I updated the logic to allow adjust the batch-size, PTAL again, thanks~

glorv:
friendly ping @sticnarf @BusyJay

Cargo.toml
```diff
@@ -9,7 +9,9 @@ license = "Apache-2.0"
 repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
-crossbeam-deque = "0.8"
+# TODO: switch back to stable version when the next version is released.
+crossbeam-deque = {git = "https://github.com/crossbeam-rs/crossbeam", rev="41ed3d948720f26149b2ebeaf58fe8a193134056"}
```
BusyJay: You should use patch instead of hard coded hash.

glorv: fixed

glorv:
@sticnarf @BusyJay  PTAL

src/queue/multilevel.rs
```diff
@@ -55,6 +55,13 @@ const MAX_LEVEL0_CHANCE: f64 = 0.98;
 /// The amount that the level 0 chance is increased or decreased each time.
 const ADJUST_AMOUNT: f64 = 0.06;
 
+/// the default max number of tasks the can steal from each level queue.
+const DEFAULT_STEAL_LIMIT_PER_LEVEL: [usize; LEVEL_NUM] = [64, 16, 1];
```
BusyJay, line 59: Are these general numbers? Or are they just optimized for copr?

glorv: I think they are general numbers. In the most common workload, the queue is not very long, and the average batch-size of each steal is about 1. So 16 should be big enough.

src/queue/multilevel.rs
```diff
@@ -384,6 +401,32 @@ impl LevelManager {
...
+
+        let cur_total_tasks_per_level: [u64; LEVEL_NUM] =
+            array::from_fn(|i| self.task_poll_duration[i].get_sample_count());
+        let cur_total_tasks = cur_total_tasks_per_level.iter().sum::<u64>();
+        let last_level0_total_tasks = self.last_exec_tasks_per_level[0].get();
+        let last_total_tasks: u64 = self.last_exec_tasks_per_level.iter().map(|c| c.get()).sum();
+
+        let level_0_tasks = (cur_total_tasks_per_level[0] - last_level0_total_tasks) as usize;
+        let total_tasks = (cur_total_tasks - last_total_tasks) as usize;
+        if total_tasks > ADJUST_LEVEL_STEAL_SIZE_THRESHOLD {
```
sticnarf, line 414: Sorry, how to understand this threshold? `maybe_adjust_chance` is called every second. So, the adjustment only happens when there are more than 100 tasks added in the last second?

sticnarf:
Yes, the steal batch size calculation is very similar to adjusting the level 0 chance. So, I'm also thinking if there is a simple empirical formula with the level 0 chance as the input 😂

src/queue/multilevel.rs
glorv: When there are not many tasks, we don't need to adjust the threshold as every steal can only pop 1 or 2 tasks. I moved the update current count of each level within the if block to make it more reasonable.

sticnarf: I got your idea somewhat. What if there are only heavy tasks and the task incoming rate is slow? For example, there are only 50 tasks coming per second but each may run for 1 second. They will all fall in the bottom level but not trigger the adjustment (if I understand correctly)

glorv: Yes. So I changed the logic to do not update the accumulated value when the total task is under the threshold, thus in this example, it can changed the batch-size after two round of adjust. PTAL again @sticnarf

sticnarf (approved):

glorv:
@BusyJay PTAL

BusyJay (approved):

PR #71, opened 2022-11-22
fix task total execute time tracker

glorv:
Fix the bug that the task total execution time is not updated on time. The bug is introduced by #67.
This bug can affect the performance impact of big task for small tasks.
I test the performace impact by running sysbench oltp_read_only(--thread=50 --range-selects) as small tasks and run go-tpc tpch --sf=10 as big tasks.
Before: 
before

After
after

The throughput impact is much better and the tail latency is slightly better.

### DIFF
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -274,6 +274,9 @@ where
         let elapsed = begin.elapsed();
 
         task_running_time.inc_by(elapsed);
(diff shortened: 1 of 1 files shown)


glorv:
@BusyJay @sticnarf  PTAL, thanks

sticnarf (approved):

PR #72, opened 2022-12-02
queue: support priority queue

glorv:
this PR introduces a new type of queue: priority queue. The priority queue sorts task by their priority value that can allow inserting tasks in the front or middle, so we can support a more flexible schedule policy.

### Why Priority

The default multi-level queue is suitable for most FIFO scenarios. But in cases that need more fine-grained control of the task schedule order, FIFO is not good enough. For example, in a shared environment for multiple users, the scheduler should schedule tasks from different users according to their share. In this case, a priority-based scheduler based on the total execution time by the user may be a workable solution.

### Drawbacks

The priority queue is not perfect for all cases. The scheduling overhead is a little bigger than the multi-level queue. And there is no default implementation that is suitable for common cases.  The schedule also doesn't implement the multi-level strategy that automatically degrades slow tasks. 

### Implement detail

The priority queue is backed by crossbeam-skiplist, the key is the priority value and the value is the task. The implementation attaches the key with an extra auto-incremented sequence number to ensure the key never duplicate with each other.

The user should provide an implementation of the trait `TaskPriorityPriovider` to generate the priority value of each task. 

### Benchmark Result

I test the priority queue performance by comparing the overall performance of tidb cluster with sysbench oltp_read_only and select_random_ranges workload.  
In this brief benchmark, the priority scheduler implements the mclock algorithm to support fair scheduling among multi resource groups. Becuase by default there is only one resource group, so it can be treat as a FIFO queue. The priority test code is: https://github.com/glorv/tikv/tree/qos-http and the nightly code is commit `2704588c6aaa1a269bb91499229e358d23cc636b`.

The benchmark is running on 3 * 8core tikv and 6 * 8 core tidb so tikv should be the performance bottleneck.
Result:

图片
(the first is priority queue and second is yatp multi-level queue)

The benchmark shows that the overhead of priority-queue is bigger than multi-level queue. The performance drops about 5% in oltp_read_only and 1% in select_random_ranges.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
 crossbeam-deque = "0.8"
-crossbeam-skiplist = { git = "https://github.com/crossbeam-rs/crossbeam" }
+crossbeam-skiplist = "0.1"
 dashmap = "5.1"
 fail = "0.5"
 lazy_static = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ repository = "https://github.com/tikv/yatp/"
 [dependencies]
 crossbeam-deque = "0.8"
 crossbeam-skiplist = "0.1"
+crossbeam-utils = "0.8"
 dashmap = "5.1"
 fail = "0.5"
 lazy_static = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
 crossbeam-deque = "0.8"
+crossbeam-skiplist = { git = "https://github.com/crossbeam-rs/crossbeam" }
 dashmap = "5.1"
 fail = "0.5"
 lazy_static = "1"
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -3,7 +3,7 @@
 use crate::pool::spawn::QueueCore;
 use crate::pool::worker::WorkerThread;
 use crate::pool::{CloneRunnerBuilder, Local, Remote, Runner, RunnerBuilder, ThreadPool};
-use crate::queue::{self, multilevel, LocalQueue, QueueType, TaskCell};
+use crate::queue::{self, multilevel, priority, LocalQueue, QueueType, TaskCell};
 use crate::task::{callback, future};
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
@@ -288,6 +288,17 @@ impl Builder {
         let queue_builder = multilevel::Builder::new(multilevel::Config::default());
         let runner_builder = queue_builder.runner_builder(fb);
         self.build_with_queue_and_runner(QueueType::Multilevel(queue_builder), runner_builder)
+    }
+
+    ///
+    pub fn build_priority_future_pool(
+        &self,
+        priority_priovider: Arc<dyn priority::TaskPriorityProvider>,
+    ) -> ThreadPool<future::TaskCell> {
+        let fb = CloneRunnerBuilder(future::Runner::default());
+        let queue_builder = priority::Builder::new(priority::Config::default(), priority_priovider);
+        let runner_builder = queue_builder.runner_builder(fb);
+        self.build_with_queue_and_runner(QueueType::Priority(queue_builder), runner_builder)
     }
 
     /// Spawns the thread pool immediately.
--- a/src/pool/builder.rs
+++ b/src/pool/builder.rs
@@ -290,13 +290,15 @@ impl Builder {
         self.build_with_queue_and_runner(QueueType::Multilevel(queue_builder), runner_builder)
     }
 
+    /// Spawn a priority future pool.
     ///
+    /// It setups the pool with priority queue. Caller must provide a `TaskPriorityProvider` implementation to generate the proper priority value for each spawned task.
     pub fn build_priority_future_pool(
         &self,
-        priority_priovider: Arc<dyn priority::TaskPriorityProvider>,
+        priority_provider: Arc<dyn priority::TaskPriorityProvider>,
     ) -> ThreadPool<future::TaskCell> {
         let fb = CloneRunnerBuilder(future::Runner::default());
-        let queue_builder = priority::Builder::new(priority::Config::default(), priority_priovider);
+        let queue_builder = priority::Builder::new(priority::Config::default(), priority_provider);
         let runner_builder = queue_builder.runner_builder(fb);
         self.build_with_queue_and_runner(QueueType::Priority(queue_builder), runner_builder)
     }
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -9,6 +9,7 @@
 //! data structs.
 
 pub mod multilevel;
+pub mod priority;
 
 mod extras;
 mod single_level;
@@ -18,7 +19,7 @@ pub use self::extras::Extras;
 use std::time::Instant;
 
 /// A cell containing a task and needed extra information.
-pub trait TaskCell {
+pub trait TaskCell: 'static {
     /// Gets mutable extra information.
     fn mut_extras(&mut self) -> &mut Extras;
 }
@@ -42,6 +43,7 @@ pub(crate) struct TaskInjector<T>(InjectorInner<T>);
 enum InjectorInner<T> {
     SingleLevel(single_level::TaskInjector<T>),
     Multilevel(multilevel::TaskInjector<T>),
+    Priority(priority::TaskInjector<T>),
 }
 
 impl<T: TaskCell + Send> TaskInjector<T> {
@@ -50,6 +52,7 @@ impl<T: TaskCell + Send> TaskInjector<T> {
         match &self.0 {
             InjectorInner::SingleLevel(q) => q.push(task_cell),
             InjectorInner::Multilevel(q) => q.push(task_cell),
+            InjectorInner::Priority(q) => q.push(task_cell),
         }
     }
 
@@ -57,6 +60,7 @@ impl<T: TaskCell + Send> TaskInjector<T> {
         match self.0 {
             InjectorInner::SingleLevel(_) => Extras::single_level(),
             InjectorInner::Multilevel(_) => Extras::multilevel_default(),
+            InjectorInner::Priority(_) => Extras::single_level(),
         }
     }
 }
@@ -80,6 +84,7 @@ pub(crate) struct LocalQueue<T>(LocalQueueInner<T>);
 enum LocalQueueInner<T> {
     SingleLevel(single_level::LocalQueue<T>),
     Multilevel(multilevel::LocalQueue<T>),
+    Priority(priority::LocalQueue<T>),
 }
 
 impl<T: TaskCell + Send> LocalQueue<T> {
@@ -88,6 +93,7 @@ impl<T: TaskCell + Send> LocalQueue<T> {
         match &mut self.0 {
             LocalQueueInner::SingleLevel(q) => q.push(task_cell),
             LocalQueueInner::Multilevel(q) => q.push(task_cell),
+            LocalQueueInner::Priority(q) => q.push(task_cell),
         }
     }
 
@@ -97,6 +103,7 @@ impl<T: TaskCell + Send> LocalQueue<T> {
         match &mut self.0 {
             LocalQueueInner::SingleLevel(q) => q.pop(),
             LocalQueueInner::Multilevel(q) => q.pop(),
+            LocalQueueInner::Priority(q) => q.pop(),
         }
     }
 
@@ -104,6 +111,7 @@ impl<T: TaskCell + Send> LocalQueue<T> {
         match self.0 {
             LocalQueueInner::SingleLevel(_) => Extras::single_level(),
             LocalQueueInner::Multilevel(_) => Extras::multilevel_default(),
+            LocalQueueInner::Priority(_) => Extras::single_level(),
         }
     }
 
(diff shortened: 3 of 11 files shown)


glorv:
@Connor1996 @sticnarf @BusyJay  PTAL

BusyJay:
Can you highlight the cases that priority queue solves better than multi level queue? And what's the drawbacks of priority queue?

glorv:
> Can you highlight the cases that priority queue solves better than multi level queue? And what's the drawbacks of priority queue?

@BusyJay  I have changed the description, PTAL again, thanks

glorv:
@Connor1996 @sticnarf  PTAL, thanks

Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/tikv/yatp/"
 
 [dependencies]
 crossbeam-deque = "0.8"
+crossbeam-skiplist = { git = "https://github.com/crossbeam-rs/crossbeam" }
```
sticnarf: crossbeam-skiplist 0.1 recently releases. We can use the crates.io version now.

src/queue/priority.rs
```diff
@@ -0,0 +1,501 @@
...
+            .map(|e| into_pop(e.value().take().unwrap()))
+    }
+
+    #[inline]
+    fn gen_key(&self, weight: u64) -> MapKey {
+        MapKey(weight, self.sequence.fetch_add(1, Ordering::Relaxed))
+    }
+}
+
+/// A holder to store task. The slot can be concurrently visit by multiple thread in the skip-list.
```
sticnarf: I only see `push` and `pop` access the skiplist and they don't seem to visit a slot concurrently. Where does concurrent visit happen?

src/queue/priority.rs
```diff
@@ -0,0 +1,501 @@
...
+/// A trait used to generate priority value for each task.
+pub trait TaskPriorityProvider: Send + Sync + 'static {
+    /// Return a priority value of this task, all tasks in the priority queue is ordered by this value.
+    fn get_priority(&self, extras: &Extras) -> u64;
+}
+
+///
+#[derive(Clone)]
+struct PriorityTaskManager {
+    level_manager: Arc<TaskLevelManager>,
```
sticnarf, line 87: With a priority queue, why is there still a level manager?

src/queue/priority.rs
glorv: Yes, in our case it's safe to take the value. But the pop() api only provide readonly access to the key and value. So here we wrap it in a atomic ptr to workaround this restriction. Do you have any idea to avoid this?

src/queue/priority.rs
glorv: While using the priority queue, we still need to auto demote slow tasks. So the priority manager can use the level information to set different priority value. Or we have to implement similar mechanism to do so.

src/queue/priority.rs
sticnarf: Ah, I see. The skiplist cannot guarantee that the entry is not borrowed by others when popped, so we only have read only access to the value.

Instead of an `AtomicPtr` wrapping a `Box`, do you think a crossbeam `AtomicCell` is simpler and cheaper? There will be no extra allocation. The future `TaskCell` is just a pointer so it should also be lock free.

glorv: Seems https://github.com/crossbeam-rs/crossbeam/blob/0de660ec88481ff7787374efee0c2e21d4d403b2/crossbeam-utils/src/atomic/atomic_cell.rs#L1020 AtomicCell is also not very cheap because it uses a global lock for large value (> 8 bytes).

glorv: Ok. I found we can simply replace the AtomicPtr<T> with `RefCell<Option<T>>`. Since only the thread pops the entry can visit the value, it's safe to impl Sync for the `Slot` type. @sticnarf  PTAL again.

sticnarf: The future taskcell is only 8 bytes so atomic cell is cheap. For other types, yes, a `Mutex` is better than the implementation of `AtomicCell`.

While I don't think `RefCell` + `unsafe impl Sync` is a good combination.

If you care about safety, `Mutex<Option<>>` is your choice. And it's fast in practice when there is not any race (only need a very few atomic operations).
If you are confident enough, just use `UnsafeCell`. `RefCell` does not provide you with extra safety when you mark it `Sync` because its borrow checks will become unreliable when multiple threads access it concurrently.

glorv: Done. Change to using `AtomicCell`. Though `UnsafeCell` is also an available choice, since this isn't a performance bottleneck, using `AtomicCell` can avoid unsafe code.

sticnarf (dismissed):
LGTM.
@BusyJay Do you want to take another look?

src/pool/builder.rs
```diff
@@ -290,6 +290,17 @@ impl Builder {
         self.build_with_queue_and_runner(QueueType::Multilevel(queue_builder), runner_builder)
     }
 
+    ///
```
Connor1996, line 294: comment

src/queue/extras.rs
```diff
@@ -28,6 +28,8 @@ pub struct Extras {
     pub(crate) fixed_level: Option<u8>,
     /// Number of execute times
     pub(crate) exec_times: u32,
+    /// The task group id. Used in priority queue.
+    pub(crate) group_id: u64,
```
Connor1996: why not group name?

glorv: Done. But I think it a bit weird here because the group_id/name is only fits in some specified cases(e.g. in tikv), but may not fit for other cases. I'm not sure if it is better to be defined as `extra_data: Vec<u8>`, so the user can put arbitary data in this field as they need.

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+            }
+        }
+
+        self.pq
+            .pop_front()
+            .map(|e| into_pop(e.value().take().unwrap()))
+    }
+
+    #[inline]
+    fn gen_key(&self, weight: u64) -> MapKey {
```
nolouch: may s/weight/prioriy/ more unify.

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+}
+
+/// The configurations of priority task queues.
+pub struct Config {
+    name: Option<String>,
+    level_time_threshold: [Duration; LEVEL_NUM - 1],
+}
+
+impl Config {
+    /// Sets the name of the multilevel task queue. Metrics of multilevel
```
nolouch:
```suggestion
    /// Sets the name of the priority task queue. 
```

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+
+/// The builder of a priority task queue.
+pub struct Builder {
+    manager: PriorityTaskManager,
+    level0_elapsed_us: IntCounter,
+    total_elapsed_us: IntCounter,
+}
+
+impl Builder {
+    /// Creates a multilevel task queue builder with specified config and [`TaskPriorityProvider`].
```
nolouch:
```suggestion
    /// Creates a prioriy task queue builder with specified config and [`TaskPriorityProvider`].
```

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+            local_level0_elapsed_us: self.level0_elapsed_us.local(),
+            local_total_elapsed_us: self.total_elapsed_us.local(),
+        }
+    }
+}
+
+/// The configurations of priority task queues.
+pub struct Config {
+    name: Option<String>,
+    level_time_threshold: [Duration; LEVEL_NUM - 1],
```
nolouch, line 255: can't configure this field?

src/queue.rs
```diff
@@ -18,7 +19,7 @@ pub use self::extras::Extras;
 use std::time::Instant;
 
 /// A cell containing a task and needed extra information.
-pub trait TaskCell {
+pub trait TaskCell: 'static {
```
BusyJay, line 22: Why?

src/queue/extras.rs
```diff
@@ -28,6 +28,8 @@ pub struct Extras {
     pub(crate) fixed_level: Option<u8>,
     /// Number of execute times
     pub(crate) exec_times: u32,
+    /// The task group name. Used in priority queue.
+    pub(crate) group_name: String,
```
BusyJay: I don't see it's set or get in any place.

src/pool/builder.rs
```diff
@@ -290,6 +290,17 @@ impl Builder {
...
+    pub fn build_priority_future_pool(
+        &self,
+        priority_priovider: Arc<dyn priority::TaskPriorityProvider>,
```
Connor1996: priovider -> provider

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+        let level = extras.current_level();
+        let total_running_time = extras.total_running_time.clone();
+        let begin = Instant::now();
+        let res = self.inner.handle(local, task_cell);
+        let elapsed = begin.elapsed();
+        let elapsed_us = elapsed.as_micros() as u64;
+        if let Some(ref running_time) = total_running_time {
+            running_time.inc_by(elapsed);
+        }
+        if level == 0 {
```
Connor1996, line 201: shouldn't always be 0?

src/queue/priority.rs
```diff
@@ -0,0 +1,492 @@
...
+    }
+}
+
+/// A trait used to generate priority value for each task.
+pub trait TaskPriorityProvider: Send + Sync + 'static {
+    /// Return a priority value of this task, all tasks in the priority queue is ordered by this value.
+    fn get_priority(&self, extras: &Extras) -> u64;
+}
+
+///
```
Connor1996: comment

src/queue/priority.rs
glorv: It's the same as multi-level, not configurable currently

src/queue.rs
glorv: The skiplist's value must be 'static. I think it's fair that a task spawned in the thread pool should live long enough.

src/queue/priority.rs
glorv: No. The priority runner also implements the level degradation logic like the multilevel queue, so the user may assign a different priority for higher level tasks.

src/queue/extras.rs
glorv: In the multi-tenant implementation, we need to attach the task with the resource group name. I change this field to `metadata: Vec<u8>` to allow priority queue user to store arbitrary data here.

glorv:
@BusyJay @Connor1996 @nolouch PTAL again, thank you very much.

src/queue.rs
BusyJay: I don't see why skiplist's value should be 'static. Technically, as long as the value outlive skiplist, it should be OK. If that's an API shortcoming, then the constraint should be put into priority queue instead of here.

src/queue/extras.rs
```diff
@@ -89,4 +94,19 @@ impl Extras {
...
+    pub fn metadata(&self) -> &[u8] {
+        &self.metadata
+    }
+
+    /// Gets the mutable metadata of this task.
+    pub fn metadata_mut(&mut self) -> &mut Vec<u8> {
+        &mut self.metadata
+    }
+
+    /// Set the group name of this task.
```
Connor1996: outdated comment

src/queue.rs
sticnarf: It looks like an API shortcoming. I tried making the constraint priority-queue-only, but because we use `enum` to dispatch instead of trait, the constraint propagates up.

src/queue/priority.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// Pushes the task cell to the queue. The schedule time in the extras is
+    /// assigned to be now.
+    pub fn push(&self, mut task_cell: T) {
+        let priority = self.task_manager.prepare_before_push(&mut task_cell);
+        set_schedule_time(&mut task_cell);
+        self.queue.push(task_cell, priority);
+    }
+}
+
+/// The local queue is just a proxy of the global queue.
```
BusyJay: Then why not just use the same struct? You can define different alias.

src/queue/priority.rs
```diff
@@ -0,0 +1,491 @@
...
+    pub(super) fn has_tasks_or_pull(&mut self) -> bool {
+        !self.queue.is_empty()
+    }
+}
+
+/// A trait used to generate priority value for each task.
+pub trait TaskPriorityProvider: Send + Sync + 'static {
+    /// Return a priority value of this task, all tasks in the priority
+    /// queue is ordered by this value.
+    fn get_priority(&self, extras: &Extras) -> u64;
```
BusyJay:
```suggestion
    fn priority_of(&self, extras: &Extras) -> u64;
```

BusyJay (commented):
How about updating benches to include microbench of the new queue?

Rest LGTM

nolouch (commented):
lgtm

BornChanger (dismissed):

Connor1996 (dismissed):
LGTM

glorv:
@BusyJay  PTAL again

BornChanger (approved):
LGTM

Connor1996 (approved):

glorv:
The failed case is caused by the extra check by Miri, and is fixed by this PR https://github.com/crossbeam-rs/crossbeam/pull/940. Since it doesn't affect correctness, we can fix in the next release version of crossbeam-skiplist.

glorv:
> The failed case is caused by the extra check by Miri, and is fixed by this PR [crossbeam-rs/crossbeam#940](https://github.com/crossbeam-rs/crossbeam/pull/940). Since it doesn't affect correctness, we can fix in the next release version of crossbeam-skiplist.

crossbeam-skip already released v0.1.1 which included the fixed PR, so we can just rerun the failed case.

PR #73, opened 2023-02-10
queue: make the cleanup interval configurable

zyguan:
https://github.com/tikv/tikv/issues/14188 shows that clearing task elapsed map may introduce tail latency. In order to resolve the issue, this PR tries to make the cleanup interval configurable and export the cleanup function to users. The auto cleanup can be disabled by setting the interval to `None`, then we can call the cleanup function manually in a background task.

The PR also fix a potential issue that the thread-local `RECENT_NOW` may be not updated properly when using priority queue.

### DIFF
--- a/src/queue/multilevel.rs
+++ b/src/queue/multilevel.rs
@@ -11,7 +11,7 @@ use crate::metrics::*;
 use crate::pool::{Local, Runner, RunnerBuilder};
 
 use crossbeam_deque::{Injector, Steal, Stealer, Worker};
-use dashmap::DashMap;
+use dashmap::{try_result::TryResult::Present, DashMap};
 use fail::fail_point;
 use prometheus::local::{LocalHistogram, LocalIntCounter};
 use prometheus::{Gauge, Histogram, HistogramOpts, IntCounter};
@@ -349,7 +349,7 @@ impl LevelManager {
         T: TaskCell,
     {
         self.task_level_mgr.adjust_task_level(task_cell);
-        task_cell.mut_extras().schedule_time = Some(self.task_level_mgr.now());
+        task_cell.mut_extras().schedule_time = Some(now());
     }
 
     fn maybe_adjust_chance(&self) {
@@ -456,15 +456,6 @@ impl TaskLevelManager {
         extras.current_level = current_level;
     }
 
-    pub(super) fn now(&self) -> Instant {
-        if self.task_elapsed_map.cleanup_interval.is_some() {
-            now()
-        } else {
-            // we do not need to update tls_recent_now if auto cleanup is disabled.
-            Instant::now()
-        }
-    }
-
     pub(super) fn try_cleanup(&self) -> Option<Instant> {
         self.task_elapsed_map.try_cleanup()
     }
@@ -540,15 +531,16 @@ impl TaskElapsedMap {
             return v.clone();
         }
         fail_point!("between-read-new-and-read-old");
-        let elapsed = match old_map.get(&key) {
-            Some(v) => {
+        let elapsed = match old_map.try_get(&key) {
+            Present(v) => {
                 let v2 = v.clone();
                 drop(v);
                 fail_point!("between-get-from-old-and-insert-into-new");
                 new_map.insert(key, v2.clone());
                 v2
             }
-            None => {
+            _ => {
+                // the key is absent or the old map is under clearing
                 fail_point!("before-insert-new");
                 let v = new_map.entry(key).or_default();
                 v.clone()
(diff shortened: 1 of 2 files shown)


src/queue/multilevel.rs
```diff
@@ -451,6 +455,23 @@ impl TaskLevelManager {
         };
         extras.current_level = current_level;
     }
+
+    pub(super) fn now(&self) -> Instant {
```
sticnarf: Maybe not so necessary? It almost costs nothing to set a TLS variable.

src/queue/multilevel.rs
```diff
@@ -533,43 +554,48 @@ impl TaskElapsedMap {
...
+        self.maybe_cleanup();
         elapsed
     }
 
+    fn try_cleanup(&self) -> Option<Instant> {
+        self.cleaning_up
+            .compare_exchange(false, true, SeqCst, SeqCst)
+            .map(|_| {
+                let old_index = self.new_index.load(SeqCst) ^ 1;
+                self.maps[old_index].clear();
```
sticnarf, line 558: `clear` acquires locks inside the map. And `get_elapsed` reads the old map.

So, I'm thinking if we want to avoid tail latency completely, maybe it's better to `swap` a new map in.

src/queue/multilevel.rs
zyguan: :joy: I'm not sure actually. I just see priority queue uses `Instant::now` directly. Should we always use `now` for simplicity?

src/queue/multilevel.rs
zyguan: Hmm, let me test it further.

src/queue/multilevel.rs
sticnarf: I think it's ok to just always use `now()`.

sticnarf:
```rust
#![feature(test)]

extern crate test;

use std::{cell::Cell, hint::black_box, time::Instant};

use test::Bencher;

thread_local!(static RECENT_NOW: Cell<Instant> = Cell::new(Instant::now()));

pub fn now_tls() -> Instant {
    let res = Instant::now();
    RECENT_NOW.with(|r| r.set(res));
    res
}

pub fn now() -> Instant {
    Instant::now()
}

#[bench]
fn bench_now(b: &mut Bencher) {
    b.iter(|| {
        black_box(now());
    });
}

#[bench]
fn bench_now_tls(b: &mut Bencher) {
    b.iter(|| {
        black_box(now_tls());
    });
}
```

The two versions have less than 1ns difference.

src/queue/multilevel.rs
glorv: We can't avoid the cost by swapping a new map because the cost of Drop(map) is almost the same as  map.clear, drop each entry(Arc::drop) is the main cost.

glorv: I see a lot of tasks spawned by tikv side just use a random task id, e.g. point_get tasks, so maybe we can avoid putting a track entry for these tasks as there are totally useless.

sticnarf: > We can't avoid the cost by swapping a new map because the cost of Drop(map) is almost the same as map.clear, drop each entry(Arc::drop) is the main cost.

I think the point of this PR is to move map cleaning out of the `spawn` routine. For example, `try_cleanup` can be called at intervals in an external background thread.

sticnarf: > I see a lot of tasks spawned by tikv side just use a random task id, e.g. point_get tasks, so maybe we can avoid putting a track entry for these tasks as there are totally useless.

This is a good point. Yet I think this PR is still valuable considering there are also many cases of explicit transactions.

glorv: I'm not against this PR, I think it can help reduce the maximum latency a bit. I just want to point out that the cost of clean up can't be totally avoided.

sticnarf: > I'm not against this PR, I think it can help reduce the maximum latency a bit. I just want to point out that the cost of clean up can't be totally avoided.

The amortized cost of clean up is tiny. So, actually we don't care about the cost. Maximum latency is the only problem we want to solve here.

If we can put in a new map and mark the old one as stale, it's possible for us to avoid any lock contension with clean up in `get_elapsed`.

glorv: I found we can just change the current `old_map.get(&key)` to `old_map.try_get(&key)` to avoid this lock contension so if the old_map is still under cleaning, we just insert a new entry to the new_map instead. Because if the old_map is under cleaning(wlock), waiting for it is meanless because we are sure to get None after the wlock is released.

sticnarf: Oh, I didn't know that. It's definitely better than swapping.

sticnarf (approved):
LGTM

glorv (commented):
LGTM

glorv (approved):

sticnarf:
Thank you @zyguan for the investigation and contribution and thank @glorv for your great comments.

PR #75, opened 2023-03-03
build: update corssbeam-deque to release versoin

glorv:
As the crossbeam-deque v0.8.3 was published, we don't need this patch anymore.


glorv:
@BusyJay @sticnarf  PTAL, thanks~

BusyJay (approved):

sticnarf:
Is it better to change crossbeam-deque to `0.8.3` because we need the new version for the new API?

PR #77, opened 2023-03-15
add task metrics for priority runner

glorv:
The different between multilevel and priority is the queuing strategy and the runner is almost the same, the only difference is that multilevel runner adjusts level 0 chance dynamically. So this PR extract the common part of the runner into a separate struct.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -58,8 +58,10 @@ impl<T: TaskCell + Send> TaskInjector<T> {
 
     pub fn default_extras(&self) -> Extras {
         match self.0 {
-            InjectorInner::SingleLevel(_) | InjectorInner::Priority(_) => Extras::single_level(),
-            InjectorInner::Multilevel(_) => Extras::multilevel_default(),
+            InjectorInner::SingleLevel(_) => Extras::single_level(),
+            InjectorInner::Multilevel(_) | InjectorInner::Priority(_) => {
+                Extras::multilevel_default()
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


glorv:
@BusyJay @Connor1996  PTAL

Connor1996 (dismissed):
LGTM

src/queue.rs
```diff
@@ -60,7 +60,7 @@ impl<T: TaskCell + Send> TaskInjector<T> {
         match self.0 {
             InjectorInner::SingleLevel(_) => Extras::single_level(),
             InjectorInner::Multilevel(_) => Extras::multilevel_default(),
-            InjectorInner::Priority(_) => Extras::single_level(),
+            InjectorInner::Priority(_) => Extras::multilevel_default(),
```
BusyJay: Better merge the same branch.

src/queue/multilevel.rs
```diff
@@ -303,6 +293,142 @@ where
...
+
+    fn start(&mut self, local: &mut Local<T>) {
+        self.inner.start(local)
+    }
+
+    fn handle(&mut self, local: &mut Local<T>, task_cell: T) -> bool {
+        let res = self.inner.handle(local, task_cell);
+        let local_total = self.inner.local_total_elapsed_us.get();
+        if local_total > FLUSH_LOCAL_THRESHOLD_US {
             self.manager.maybe_adjust_chance();
```
BusyJay, line 450: This will be unreachable because `local_total_elapsed_us` is flushed.

glorv: fixed, PTAL again

BusyJay (approved):

Connor1996 (approved):

PR #78, opened 2023-03-17
Make prioirty tuple

Connor1996:
to support scheduling based on group priority

### DIFF
--- a/src/queue/priority.rs
+++ b/src/queue/priority.rs
@@ -71,7 +71,7 @@ where
 }
 
 /// priority key
-#[derive(PartialEq, PartialOrd, Eq, Ord, Default)]
+#[derive(PartialEq, Copy, PartialOrd, Eq, Ord, Default)]
 pub struct Priority {
     ///
     pub priority: u64,
--- a/src/queue/priority.rs
+++ b/src/queue/priority.rs
@@ -70,13 +70,17 @@ where
     }
 }
 
-/// priority key
+/// Priority represents the priority of a task.
 #[derive(PartialEq, Copy, Clone, PartialOrd, Eq, Ord, Default)]
 pub struct Priority {
-    ///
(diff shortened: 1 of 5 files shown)


src/queue/priority.rs
```diff
@@ -70,11 +70,39 @@ where
...
+            .then_with(|| self.value.cmp(&other.value))
+    }
+}
+
 /// A trait used to generate priority value for each task.
 pub trait TaskPriorityProvider: Send + Sync + 'static {
     /// Return a priority value of this task, all tasks in the priority
     /// queue is ordered by this value.
-    fn priority_of(&self, extras: &Extras) -> u64;
+    fn priority_of(&self, extras: &Extras) -> Priority;
```
glorv, line 105: Typically, the total number of priorities is not very big, so what about sharing the high bits of the value field, so we can still use a u64 here and keep this trait simple.

Connor1996: I don't think it makes the trait complicated.  A defined struct is more readable and extensible

glorv: But not all case need the `task_priority`, this change some what put implementation details into this trait

src/queue/priority.rs
```diff
@@ -70,11 +70,39 @@ where
...
+    fn default() -> Self {
+        Priority {
+            group_priority: 8,
+            value: 0,
+        }
+    }
+}
+
+impl Ord for Priority {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
```
BusyJay, line 93: What's the difference from `#[derive(Ord)]`?

Connor1996: note that `group_prioirty` is in descending order

Connor1996:
Close it as we plan to do it TiKV by encoding the priority as a u64

PR #88, opened 2026-02-14
Updated QueueCore#ensure_workers to reduce the runtime complexity

gaojieliu:
This function is calling parking_lot_core::unpark_filter to unpark sleeping threads and every time, it would only unpark one thread, but the original code will iterate all the parked threads regardless. Here is the original discussion:
https://github.com/tikv/yatp/pull/65/changes#r809711564 The reason is that we need to take care of the scenario, where the core_thread_count is smaller than max_thread_count, so the PR owner chooses to use FilterOp::Skip to scan all the parked threads all the time, which is very inefficient.
To handle all the scenarios, this PR will check whether it has already unparked one thread or not, if yes, return FilterOp::Stop, otherwise, return FilterOp::Skip. This will reduce the runtime complexity from O(n) [n: parked thread] to O(1) when core_thread_count is equal to max_thread_count.
This is important since this logic is in hot path (push) and the whole unpark logic is guarded by a global lock per queue.

## Summary by CodeRabbit

* **Chores**
  * Wake-up scanning in the thread pool now stops early once sufficient threads have been woken, reducing unnecessary scanning and CPU overhead during wake-up cycles.
* **New Features**
  * Local work queues now expose a public numeric identifier.
* **Tests**
  * Added a test validating wake/unpark behavior across multiple ensure-worker invocations and orderly thread wakeups.

### DIFF
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -73,6 +73,9 @@ impl<T> QueueCore<T> {
                     {
                         unparked_once = true;
                         FilterOp::Unpark
+                    } else if unparked_once {
+                        // We have unparked enough threads, so we can stop scanning the queue to save time.
+                        FilterOp::Stop
                     } else {
                         FilterOp::Skip
                     }
@@ -280,6 +283,11 @@ impl<T: TaskCell + Send> Local<T> {
             local_queue,
             core,
         }
+    }
+
+    /// Get the ID of the local queue.
+    pub fn id(&self) -> usize {
+        self.id
     }
 
     /// Spawns a task to the local queue.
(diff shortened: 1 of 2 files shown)


gaojieliu:
Close it for now.

gaojieliu:
@BusyJay @ethercflow @sticnarf 
Since you folks were involved in the original conversion in this comment:
https://github.com/tikv/yatp/pull/65/changes#r809711564
Can you take a look at this change to see whether it makes sense or not?

gaojieliu:
@ethercflow 
Can you share how I can run the benchmark against this change?
I was referring to this benchmark mentioned in another PR:
https://github.com/tikv/yatp/pull/65#issuecomment-1043029817

PR #90, opened 2026-03-31
pool: fix scaled-down threads still processing tasks

mittalrishabh:
Issue [#19498](https://github.com/tikv/tikv/issues/19498)
When threads are scaled down, threads with ID > core_thread_count should go to sleep immediately. Previously, the park validation callback would still pop tasks from the queue, causing scaled-down threads to keep working instead of sleeping.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where worker threads continued processing work even after the thread pool was scaled down to a smaller number of active workers, ensuring they properly enter sleep mode when no longer needed.

### DIFF
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -329,6 +329,11 @@ impl<T: TaskCell + Send> Local<T> {
                     if !self.core.mark_sleep() {
                         return false;
                     }
+                    // If this thread is above core_thread_count, go to sleep
+                    // without popping so scaled-down threads don't keep working.
+                    if id > self.core.config.core_thread_count.load(Ordering::SeqCst) {
+                        return true;
+                    }
                     task = self.local_queue.pop();
                     task.is_none()
                 },


src/pool/spawn.rs
```diff
@@ -329,6 +329,11 @@ impl<T: TaskCell + Send> Local<T> {
                     if !self.core.mark_sleep() {
                         return false;
                     }
+                    // If this thread is above core_thread_count, go to sleep
+                    // without popping so scaled-down threads don't keep working.
+                    if id > self.core.config.core_thread_count.load(Ordering::SeqCst) {
```
glorv, line 334: Can you explain the requirement that need to let the thread park immediately instead of trying to drain the tasks in the local queue? I think  in most scenario, the difference should be very small.

mittalrishabh: We use the TiKV thread pool auto-scaling [feature](https://docs.pingcap.com/tidb/stable/tikv-configuration-file/#cpu-threshold-new-in-v855) and observed that under load, TiKV scales down the number of threads — but the yatp pool does not shrink because its threads are occupied draining the queue and unified read pool CPU remains higher than the number of active threads shown by TiKV

glorv (approved):

PR #92, opened 2026-06-25
queue: Add custom task queue support

lcwangchao:
## Summary

This PR adds support for using a user-provided task queue in YATP.

- Introduces `queue::TaskQueue<T>` and `queue::PopResult<T>` so a queue can report whether work is ready, delayed, or absent.
- Adds `Builder::build_custom_future_pool` and wires custom queues into the existing injector/local-queue plumbing.
- Extends worker parking to understand `PopResult::Pending { retry_at }`: a worker may sleep until the retry deadline, while a later `spawn` can still wake it immediately.
- Adds `task::future::yield_to_scheduler()` for future tasks that need to force the next poll through the scheduler queue instead of relying on the normal preemption hint.
- Drains custom queues on shutdown and adds coverage for pending deadlines, wake races, shutdown drain, metrics, custom pool construction, and forced scheduler yield.

## Custom Queue API

A custom queue implements `queue::TaskQueue<T>`:

```rust
pub trait TaskQueue<T>: Send + Sync + 'static {
    fn push(&self, task_cell: T);

    fn pop(&self) -> PopResult<T>;

    fn drain(&self);

    fn has_ready_task(&self) -> bool;
}
```

The most important contract is `pop()`:

- `PopResult::Ready(pop)` means a task can run immediately.
- `PopResult::Pending { retry_at }` means the queue contains tasks, but none can run yet. The worker will retry no later than `retry_at`, and can still be woken earlier by a later push.
- `PopResult::Empty` means the queue contains no tasks.

Important: if the queue still owns delayed or throttled tasks, it should return `Pending`, not `Empty`. Returning `Empty` can let workers park as if no work exists, which may delay those tasks for a long time.

## Example

Suppose a custom queue delays tasks and keeps the earliest runnable task at the top of a heap. For brevity, `Scheduled<T>` is ordered by `ready_at`, and the heap stores `Reverse<Scheduled<T>>` so the earliest time is popped first.

```rust
struct Scheduled<T> {
    task_cell: T,
    schedule_time: Instant,
    ready_at: Instant,
}

struct DelayQueue<T> {
    tasks: Mutex<BinaryHeap<Reverse<Scheduled<T>>>>,
    delay: Duration,
}
```

`push` records when the task entered the queue and when it can first run:

```rust
fn push(&self, task_cell: T) {
    let now = Instant::now();
    self.tasks.lock().unwrap().push(Reverse(Scheduled {
        task_cell,
        schedule_time: now,
        ready_at: now + self.delay,
    }));
}
```

`pop` checks the earliest scheduled task. If it is ready, the worker can run it now. If it is not ready, the queue returns the earliest retry deadline instead of pretending to be empty.

```rust
fn pop(&self) -> PopResult<T> {
    let mut tasks = self.tasks.lock().unwrap();

    let Some(Reverse(next)) = tasks.peek() else {
        return PopResult::Empty;
    };

    if Instant::now() < next.ready_at {
        return PopResult::Pending {
            retry_at: next.ready_at,
        };
    }

    let Reverse(task) = tasks.pop().unwrap();
    PopResult::Ready(Pop {
        task_cell: task.task_cell,
        schedule_time: task.schedule_time,
        from_local: false,
    })
}
```

The remaining trait methods can stay simple for this queue: `drain()` clears the heap, and `has_ready_task()` checks whether `tasks.peek()` is already ready without removing it.

A custom future pool can then be built by passing the queue as `Arc<dyn TaskQueue<future::TaskCell>>`:

```rust
let queue: Arc<dyn TaskQueue<future::TaskCell>> = Arc::new(DelayQueue::new(delay));

let mut builder = Builder::new("custom-queue");
builder.max_thread_count(4);
let pool = builder.build_custom_future_pool(queue);

pool.spawn(async move {
    // This future will be inserted into DelayQueue first, then run after the delay.
});
```

## `yield_to_scheduler`

YATP future tasks already have `task::future::reschedule()`, which gives the scheduler a chance to run other ready work. However, `reschedule()` still follows the normal preemption check: if the scheduler does not report ready work, the current worker may keep polling the same task for locality.

`task::future::yield_to_scheduler()` is stricter. It always returns the current future task to the scheduler queue before polling it again, skipping the `has_ready_task()` / preemption hint. This is useful when the queue itself should make the next scheduling decision, for example when a custom queue uses delayed readiness, rate limiting, or another policy that should be applied between task slices.

```rust
pool.spawn(async move {
    process_one_batch().await;

    // Force the next poll to go through the queue again, even if the worker
    // would otherwise continue polling this task immediately.
    yatp::task::future::yield_to_scheduler().await;

    process_next_batch().await;
});
```

## Tests

- `cargo fmt --check`
- `cargo clippy --tests -- -D clippy::all`
- `cargo clippy --tests --features failpoints -- -D clippy::all`
- `cargo test --tests`
- `cargo test --tests --features failpoints`
- `cargo test yield_to_scheduler`

## Summary by CodeRabbit

* **New Features**
  * Added support for custom task queues, including typed queue selection and building custom future pools from a provided queue.
  * Introduced a unified pop-result API (ready/pending/empty) across queue and worker handling.
* **Bug Fixes**
  * Improved worker scheduling to honor the earliest retry deadline and refine park/validate retry behavior.
  * Improved shutdown handling by draining local work more reliably.
  * Hardened steal-sizing logic to avoid divide-by-zero.
* **Tests**
  * Expanded coverage for custom queues, timing/retry behavior, shutdown boundaries, and updated tests for the new pop-result API.


src/task/future.rs
```diff
@@ -299,7 +299,7 @@ unsafe fn clone_task(task: *const ()) -> TaskCell {
 
 thread_local! {
     /// Local queue reference that is set before polling and unset after polled.
-    static LOCAL: Cell<*mut Local<TaskCell>> = Cell::new(std::ptr::null_mut());
+    static LOCAL: Cell<*mut Local<TaskCell>> = const { Cell::new(std::ptr::null_mut()) };
```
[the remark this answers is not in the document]
lcwangchao: I think this is a false positive. The `const { ... }` here is the special const initializer syntax supported by `std::thread_local!`, not general inline const expressions. Rust release notes document that this syntax was stabilized in Rust 1.59, so it does not require Rust 1.79. Since this crate does not currently declare an MSRV and CI tests stable/nightly, I don't think this PR should add `rust-version = "1.79"` or refactor this initialization.

lcwangchao:
@coderabbitai review it again

jiadebin:
Review Finding:
P1: src/pool/worker.rs:27 / src/pool/spawn.rs:348: timed Pending wakeups can bypass core_thread_count after scale-down. pop_or_sleep() only checks id > core_thread_count while validating the park. When a custom queue parks workers on Pending { retry_at }, those workers wake by timeout and pop_or_sleep() returns None; the outer worker loop then immediately starts the spin path and calls self.local.pop() without any scale-down guard. Repro: custom future queue with 4 delayed tasks, max_thread_count=4, core_thread_count=4, then scale_workers(1) before the retry deadline. At the deadline all 4 workers executed concurrently (max_in_flight=4, expected 1). This breaks the pool’s scale-down contract and is especially risky for the custom-queue/rate-limit use case this PR enables. I’d gate the spin pop for workers above the current core count, or make the timeout path re-enter the scaled-down park state instead of returning to the unguarded spin path.

src/pool/worker.rs
```diff
@@ -24,16 +24,18 @@ where
     fn pop(&mut self) -> Option<Pop<T>> {
         // Wait some time before going to sleep, which is more expensive.
         let mut spin = SpinWait::new();
-        loop {
-            if let Some(t) = self.local.pop() {
-                return Some(t);
-            }
+        let initial_retry_at = loop {
+            let retry_at = match self.local.pop() {
```
cfzjywxk, line 28: The `pop` naming is confusing before, for `retry_at` it does `top` actually.
Not related to this PR, maybe we could name it `pop_if_ready` or something more understandable.

src/pool/worker.rs
```diff
@@ -151,4 +577,393 @@ mod tests {
...
+            PopResult::Pending {
+                retry_at: earlier_retry_at,
+            },
+            ScriptedQueue::ready(1),
+        ]));
+        let mut local = build_scripted_local(queue);
+
+        assert!(local.pop_or_sleep(Some(later_retry_at)).is_none());
+        assert!(Instant::now() >= earlier_retry_at);
+        assert!(Instant::now() < later_retry_at);
```
cfzjywxk, line 716: This would be easy to be impacted by the CI environment using wall clock time.

src/pool/worker.rs
```diff
@@ -151,4 +577,393 @@ mod tests {
...
+        results.push(PopResult::Pending {
+            retry_at: earlier_retry_at,
+        });
+        results.push(PopResult::Empty);
+        results.push(ready_callback_pop(done_tx));
+        let queue = Arc::new(ScriptedQueue::new(results));
+
+        let executed_at = check_scripted_worker_runs_task(queue, done_rx);
+        assert!(executed_at >= earlier_retry_at);
+        assert!(executed_at < later_retry_at);
```
cfzjywxk, line 778: Ditto.

src/pool/worker.rs
```diff
@@ -116,6 +121,427 @@ mod tests {
...
+    ) -> callback::TaskCell {
+        callback::TaskCell {
+            task: callback::Task::new_once(task),
+            extras: Extras::multilevel_default(),
+        }
+    }
+
+    const WORKER_SPIN_POP_COUNT: usize = 11;
+    const DELAYED_TASK_DELAY: Duration = Duration::from_millis(50);
+    const MAX_DELAYED_TASK_LAG: Duration = Duration::from_millis(500);
```
cfzjywxk: Could be unstable in CI env, how about a larger value?

src/queue/custom.rs
```diff
@@ -0,0 +1,686 @@
...
+                .unwrap()
+                .get()
+                > 100_000
+        );
+        assert_eq!(
+            TASK_WAIT_DURATION
+                .get_metric_with_label_values(&[name])
+                .unwrap()
+                .get_sample_count(),
+            4
```
cfzjywxk: This would fail if it is retried with the same process. Not sure if it is a problem

src/queue/custom.rs
```diff
@@ -0,0 +1,686 @@
...
+
+        pool.spawn(async {});
+
+        // Wait until the worker has observed Pending. This makes the test check
+        // shutdown-from-pending rather than shutdown-before-the-worker-runs.
+        pending_rx.recv_timeout(Duration::from_secs(1)).unwrap();
+        assert_eq!(queue.len(), 1);
+
+        pool.shutdown();
+        assert_eq!(queue.drain_count(), 1);
```
cfzjywxk: This contracts with the trait comment `"shutdown may call this method multiple times or concurrently. Implementations must be idempotent."`

src/queue/multilevel.rs
```diff
@@ -537,15 +537,17 @@ impl LevelManager {
...
-                LEVEL_MAX_QUEUE_MAX_STEAL_SIZE
-            } else {
-                // by default level0 contains 80% of all tasks, so in the most common case, only
-                // pop 1 task from level max once, and increases level max batch size when the executed
-                // tasks are more than level0.
-                std::cmp::min(total_tasks / level_0_tasks, LEVEL_MAX_QUEUE_MAX_STEAL_SIZE)
-            };
+            // When level 0 has no tasks, the current workloads are all
+            // low-priority tasks, so use the maximum steal size.
+            let new_steal_count = total_tasks.checked_div(level_0_tasks).map_or(
```
cfzjywxk, line 542: Why change this? Seems unrelated change?

src/queue/priority.rs
```diff
@@ -673,31 +673,33 @@ mod tests {
...
             remote.spawn(MockTask::new(35, i));
         }
-        while let Some(Pop { task_cell, .. }) = locals[0].pop() {
+        while let PopResult::Ready(Pop { task_cell, .. }) = locals[0].pop() {
             assert!(runner.handle(&mut locals[0], task_cell));
         }
+        runner.flush();
 
-        // we spawn 4 tasks here but the metrics of the last one is not flush, so only check the first 3 here.
+        // Explicitly flush local metrics so the assertions do not depend on
```
cfzjywxk, line 704: Ditto seems unrelated change.

lcwangchao:
> jiadebin

Thanks for the detailed report. Fixed in a3f5419 by handling the `TimedOut + scaled-down worker` path inside `pop_or_sleep`.

When a worker above the current `core_thread_count` wakes by a pending retry timeout, it now wakes one core worker to re-check the queue and then re-parks without carrying the old timeout, instead of returning to the outer worker loop and entering the spin-pop path.

I chose this approach to keep the existing scale-down behavior for already-running workers unchanged, while fixing the new timed-Pending wakeup path introduced by custom queues. I also added failpoint tests covering:
- a scaled-down worker re-parks after a pending timeout;
- the pending deadline is handed off to a core worker, and the scaled-down worker does not execute the delayed task.

src/pool/worker.rs
lcwangchao: Good point. However, since this rename would touch the custom trait, local queue wrappers, built-in queue implementations, and many call sites, I’d prefer to keep the existing naming in this PR and handle it as a separate cleanup if we want to make the API wording more precise.

src/pool/worker.rs
lcwangchao: Good point. I relaxed these timing checks for CI stability: `MAX_DELAYED_TASK_LAG` is now 2s, `LATER_RETRY_OFFSET` is now 10s, and related `recv_timeout`s use the larger tolerance instead of 500ms. The test intent remains the same, but it should be less sensitive to wall-clock jitter.

src/pool/worker.rs
lcwangchao: Covered by the same timing relaxation.

src/pool/worker.rs
lcwangchao: Covered by the same timing relaxation. I've enlarged MAX_DELAYED_TASK_LAG to 2s

src/queue/custom.rs
lcwangchao: Fixed. The test now only requires drain to happen at least once and verifies the queue is cleared.

src/queue/multilevel.rs
lcwangchao: To make sure `cargo clippy` to pass

src/queue/priority.rs
lcwangchao: This was added because the metrics test was unstable before and hard to make pass reliably. `runner.flush()` makes the local metrics visible deterministically, so the test no longer depends on whether the internal flush threshold happens to be crossed.

src/queue/custom.rs
```diff
@@ -0,0 +1,665 @@
...
+
+impl<T: 'static> Builder<T> {
+    /// Creates the injector and local queue handles of the custom task queue.
+    pub(crate) fn build(
+        self,
+        local_num: usize,
+    ) -> (super::TaskInjector<T>, Vec<super::LocalQueue<T>>) {
+        let injector = TaskInjector::new(self.queue.clone());
+        let locals: Vec<LocalQueue<T>> =
+            std::iter::repeat_with(|| LocalQueue::new(self.queue.clone()))
```
cfzjywxk, line 105: Builder::build hands the injector and all local_num local handles the same self.queue.clone() of a single Arc<dyn TaskQueue>. Unlike the built-in multilevel/priority paths, every push/pop here contends on one shared queue.

This is amplified by `yield_to_scheduler` (src/task/future.rs:455): it re-enters `pop()` every slice unconditionally, which hits the shared lock once per slice. At high core counts under load, the slice cost may be dominated by lock contention.

Perhpas  we need to get a microbench at, say, 16/32/64 workers comparing this against the built-in queue, and confirm `MAX_SLICE` is large enough to amortize the per-slice re-queue + lock acquisition.

src/queue/custom.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    /// Pops a task from the custom queue.
+    #[inline]
+    pub fn pop(&self) -> PopResult<T> {
+        self.queue.pop()
+    }
+
+    /// Forcefully drains all tasks from the custom queue.
+    #[inline]
+    pub fn drain(&self) {
```
cfzjywxk, line 188: Better to comment that it would be called concurrectly, it must be idempotent and thread-safe, and have test case to verify this.

jiadebin (approved):

src/queue/custom.rs
lcwangchao: done

src/queue/custom.rs
lcwangchao: The key point is that the performance mainly depends on the custom queue implementation. `Arc<dyn TaskQueue>` does not require a single physical queue or a single global lock. A custom queue can still use per-worker/thread-local shards internally, or use a low-contention structure like the built-in priority queue’s skiplist-based shared queue. So I don’t think the custom API is inherently worse than the existing priority queue; the benchmark is probably more useful when TiKV adds its concrete custom queue implementation.

Also, `yield_to_scheduler()` is not intended as the default yielding primitive. The default recommendation is still `reschedule()`. It is mainly for special cases where a task must be forced back to the queue, for example after it observes that it has become rate-limited.

cfzjywxk (dismissed):

cfzjywxk (approved):

PR #93, opened 2026-07-01
queue: set schedule time for custom queue pushes

lcwangchao:
## What changed

This PR updates the custom queue adapter to set `Extras::schedule_time` before forwarding tasks through both `custom::TaskInjector::push` and `custom::LocalQueue::push`.

It also exposes read-only getters for `multilevel::Config`, so callers that build custom queues can reuse the multilevel defaults without duplicating them.

Finally, it relaxes a worker wake test to assert the scheduling guarantee instead of an exact `resume` count. The exact count is timing-sensitive because a late `ensure_workers` can wake an already-idle worker again after the task has been handled.

## Why

Built-in queues set `schedule_time` when a task is pushed, and tracked runners rely on it to report queue wait time. Custom queues already reuse the same tracked runner metrics, but their adapter did not initialize this timestamp before handing tasks to the user-provided queue.

Without this, custom queues can miss schedule-wait accounting even when the task itself uses multilevel-compatible extras.

## Tests

- `cargo fmt -- --check`
- `RUSTFLAGS="-D warnings" cargo clippy --all --all-features -- -D clippy::all`
- `RUSTFLAGS="-D warnings" cargo test --all --all-features -- --nocapture`
- `RUSTFLAGS="-D warnings" cargo bench --all --all-features -- --test`


src/queue/custom.rs
```diff
@@ -7,8 +7,8 @@
 use std::sync::Arc;
 
 use super::{
-    multilevel::{MultiLevelMetrics, TrackedRunnerBuilder},
-    PopResult,
+    multilevel::{now, MultiLevelMetrics, TrackedRunnerBuilder},
```
cfzjywxk, line 10: Should we move these crates to some common lib? Introducing `multilevel` here is a bit strange.
Non-block

cfzjywxk (approved):

src/queue/custom.rs
lcwangchao: maybe we can do it in another PR

PR #95, opened 2026-07-28
ARM64 threading optimization - atomic ordering and cache line alignment

mohakgoel1:
### Background

yatp was originally optimized primarily for x86. On ARM64, unnecessary SeqCst atomic ordering and false sharing can add performance overhead due to ARM's weaker memory model and 128-byte cache lines on server CPUs such as AWS Graviton and ARM Neoverse.

#### 1. Atomic ordering

Replaced unnecessary SeqCst operations with the minimum required memory ordering across 33 operations in 4 files:

- task/future.rs
- pool/spawn.rs
- pool/builder.rs
- queue/multilevel.rs

Used Acquire, Release, or AcqRel based on the synchronization requirements of each operation.

These changes reduce unnecessary memory barriers on ARM64 while preserving the existing synchronization behavior. No logic changes were made.

#### 2. Cache line alignment

Added a custom CacheAligned struct in pool/spawn.rs to prevent false sharing between active_workers and global_queue.

The struct uses explicit alignment:

- 128 bytes for aarch64, x86_64, and powerpc64
- 64 bytes for 32-bit arm

This ensures appropriate cache line separation across supported architectures.

### Benchmark Results

TPC-C benchmark with 1 TiKV server and 750 warehouses:

| Platform | Optimized TPMC | Base TPMC | Improvement |
|----------|----------------|-----------|-------------|
| ARM64    | 153,464        | 146,582   | +5%         |
| x86      | 169,087        | 166,125   | +2%         |

The optimization provides a larger improvement on ARM64, while also showing a smaller positive impact on x86.

### Correctness

- All 40 existing yatp unit tests pass.
- No functional or logic changes were introduced.
- Each memory ordering change was reviewed against its corresponding synchronization operation.
- ARM64 disassembly was checked to verify the expected instruction selection.

This PR is a joint contribution by:

- Abhishek Jain(@abhijain1204fujitsu)
- Mohak Goel(@mohakgoel1)

## Summary by CodeRabbit

* **Performance**
  * Reduced contention around worker activity tracking.
  * Lowered synchronization overhead during task wake-up, polling, scaling, and queue coordination.

* **Bug Fixes**
  * Improved reliability of concurrent scheduling, including worker shutdown, sleeping, waking, and task state transitions.
  * Strengthened synchronization for thread-count configuration and multilevel queue maintenance.

### DIFF
--- a/src/pool/spawn.rs
+++ b/src/pool/spawn.rs
@@ -14,11 +14,21 @@ use std::sync::{
 };
 use std::time::Instant;
 
-#[cfg(any(
-    target_arch = "x86_64",
-    target_arch = "aarch64",
-    target_arch = "powerpc64",
-))]
+// We intentionally use a custom CacheAligned struct instead of
+// crossbeam_utils::CachePadded because CachePadded aligns to only
+// 32 bytes on 32-bit arm targets, which may not fully prevent false
+// sharing on all arm32 cache line sizes. This explicit struct uses
+// platform-specific #[repr(C, align(N))] to guarantee correct
+// alignment on every target:
+//   aarch64              -> 256 bytes (conservative: covers Fujitsu A64FX 256-byte
+//                          cache lines; safe for Neoverse/Graviton 128-byte lines)
+//   x86-64, powerpc64   -> 128 bytes
+//   arm (32-bit)        -> 64 bytes
+//   all others          -> 64 bytes
+#[cfg(target_arch = "aarch64")]
+const CACHE_LINE_SIZE: usize = 256;
+
+#[cfg(any(target_arch = "x86_64", target_arch = "powerpc64"))]
 const CACHE_LINE_SIZE: usize = 128;
 
 #[cfg(target_arch = "arm")]
@@ -32,9 +42,15 @@ const CACHE_LINE_SIZE: usize = 64;
 )))]
 const CACHE_LINE_SIZE: usize = 64;
 
+#[cfg(target_arch = "aarch64")]
+#[repr(C, align(256))]
+struct CacheAligned {
+    value: AtomicUsize,
+    _pad: [u8; CACHE_LINE_SIZE - std::mem::size_of::<AtomicUsize>()],
+}
+
 #[cfg(any(
     target_arch = "x86_64",
-    target_arch = "aarch64",
     target_arch = "powerpc64",
 ))]
 #[repr(C, align(128))]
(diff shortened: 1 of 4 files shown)


src/pool/spawn.rs
```diff
@@ -36,17 +109,18 @@ pub fn is_shutdown(cnt: usize) -> bool {
...
+    active_workers: CacheAligned,
     config: SchedConfig,
 }
 
 impl<T> QueueCore<T> {
     pub fn new(global_queue: TaskInjector<T>, config: SchedConfig) -> QueueCore<T> {
         QueueCore {
             global_queue,
-            active_workers: AtomicUsize::new(config.max_thread_count << WORKER_COUNT_SHIFT),
+            active_workers: CacheAligned::new(config.max_thread_count << WORKER_COUNT_SHIFT),
```
BusyJay: How about crossbeam::utils::CachePadded?

mohakgoel1: Hi @BusyJay,
Thanks for the review and for pointing this out.

We considered crossbeam_utils::CachePadded but moved away from it for two reasons that came up earlier in this review:

- On 32-bit arm, CachePadded aligns to only 32 bytes, which may not fully prevent false sharing on all arm32 cache line sizes.
- On aarch64, CachePadded aligns to 128 bytes, but the Fujitsu A64FX a production aarch64 server chip used in supercomputing uses 256-byte cache lines. With 128-byte alignment, active_workers could still share a cache line with an adjacent QueueCore field on that platform.

The explicit #[repr(C, align(N))] struct with per-platform #[cfg] blocks gives us full control over both size and alignment on every target, covering all the above cases correctly. The comment above the struct documents this reasoning so future readers understand why CachePadded was not used.

BusyJay: Can you also send a patch to crossbeam and see whether they accept the changes?

BusyJay: If they don't, we can reuse definitions from crossbeam by reexporting them in wrapper mode as a fallback, and only define what we know to be wrong in our own crate.

mohakgoel1: Thank you for the suggestion. Our thought is to go with Option 2 the wrapper/fallback pattern feels like the right approach here. It keeps the custom logic minimal by only overriding the two platforms where crossbeam's alignment is known to be insufficient (aarch64 at 256 bytes and arm at 64 bytes), and re-exports CachePadded directly for everything else. This avoids duplicating crossbeam's full platform logic in yatp, is easier to maintain long term, and if crossbeam ever updates their alignment values upstream, removing our overrides becomes a one-line cleanup.

What are your thoughts on proceeding with Option 2?

mohakgoel1: Hi @BusyJay, Could you please confirm if we should proceed with the wrapper/fallback approach?

mohakgoel1: Hi @BusyJay, Just following up on my previous comment. Could you please confirm if we can proceed with the wrapper/fallback approach? Thanks!

mohakgoel1: Hi @BusyJay,
We've pushed a commit removing the custom CacheAligned struct and replacing it with CachePadded<AtomicUsize> for active_workers in QueueCore. On aarch64 this gives 128-byte alignment, which is safe and correct.

Let me know if this looks good to you, or if you'd like any further changes before merging.

glorv (approved):
LGTM

LykxSassinator (approved):

mohakgoel1:
Hi @BusyJay, @Connor1996, @gengliqi,
Just a gentle reminder regarding the review of this PR.

We have addressed the previous feedback and updated the PR accordingly. Could you please take a look when you get a chance? Your review and feedback would be greatly appreciated.

glorv:
Thank you @mohakgoel1 for your contribution~

mohakgoel1:
> Thank you @mohakgoel1 for your contribution~

Thank you for taking the time to review the PR and for your valuable feedback. I really appreciate your support and help in getting it merged!

PR #96, opened 2026-08-27
pool: park scaled-down workers on a queue that never drains

mittalrishabh:
### What is changed and how it works

`WorkerThread::pop` spin-pops the local queue and returns as soon as a task is ready. `pop_or_sleep` — the only place that compares the worker id against `core_thread_count` — is therefore never reached while the queue has work in it.

The result is that scaling in has no effect exactly when it is needed. Under a sustained backlog every worker keeps finding a task, so `scale_workers(n)` lowers `core_thread_count` but the surplus threads never park. We saw this on a TiKV unified read pool: the pool reported 7 core threads while measuring 8.8 cores of CPU, with 7k–12k tasks queued continuously.

This moves the check ahead of the spin-pop loop: a worker above `core_thread_count` pauses and parks straight away. Once parked, the existing logic in `pop_or_sleep`'s `validate` callback keeps it asleep until it is back within the core count, and `unpark_one`'s park-token filter already declines to wake it.

Custom queues are exempt. Only `custom::LocalQueue::pop` can return `PopResult::Pending`, and that deadline is observed on the spin-pop path, so scaled-down workers on a custom queue keep the old behaviour (`test_scaled_down_pending_timeout_wakes_core_worker` covers it). The `SingleLevel`/`Multilevel`/`Priority` queues convert through `Option<Pop<T>>` and can never produce `Pending`.

### Check List

Tests

- Unit test: `test_scaled_down_worker_parks_while_tasks_are_queued` — a worker above `core_thread_count` facing 8 queued tasks must park with `handle == 0`. Without the fix it runs all 8.

Side effects

- None expected. Behaviour changes only for workers whose id is above `core_thread_count`, which are already meant to be idle.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved worker scaling so surplus workers pause promptly when the system scales down, even while tasks remain queued.
  - Preserved timely processing for custom queues that rely on scheduled task deadlines.
  - Improved responsiveness during scale-in operations without interrupting eligible queued work.

- **Tests**
  - Added coverage verifying that scaled-down workers pause while queued tasks are still present.


v01dstar (dismissed):

src/pool/worker.rs
```diff
@@ -22,6 +22,15 @@ where
 {
     #[inline]
     fn pop(&mut self) -> Option<Pop<T>> {
+        // A surplus worker has to stop taking work, not merely stop being woken
+        // up: on a queue that never drains, the spin-pop below always finds a
+        // task and `core_thread_count` never takes effect.
+        if self.local.should_park_before_pop() {
```
glorv:
```suggestion
        if !self.local.should_park_before_pop() {
           let initial_retry_at = loop {
             ...
           }
        }
        self.runner.pause(&mut self.local);
        let t = self.local.pop_or_sleep(initial_retry_at);
        self.runner.resume(&mut self.local);
        return t;
```

glorv (commented):
rest LGTM

src/pool/worker.rs
mittalrishabh: should_park_before_pop returns true if it is a scaled down worker.

glorv: I mean `should_park_before_pop` checks whether to do the spinlock check. So you can move the spin loop into the should_park check to reduce code redundancy.

glorv (approved):

